PR #22, opened 2023-08-31
Pykeops ndft

paquiteau:
This PR add support for NDFT using PyKeops. 

We thus have a nice implementation of the NDFT that is fast, and can be launch either on GPU or CPU.


paquiteau:
Alright so pykeops if working well  on CPU as some trouble running on GPU (https://github.com/getkeops/keops/issues/328) I think we can do without it and maybe use other accelerated NDFT (using numba for instance, or cupy on gpu). Yet they were some adds to tests in the PR, so I will refactor this out and resubmit it.

PR #27, opened 2023-09-13
Add 2D ring trajectory

Daval-G:
This PR adds a 2D ring-based trajectory, initially proposed by Wu, Hochong H., Jin Hyung Lee, and Dwight G. Nishimura. “MRI Using a Concentric Rings Trajectory.” Magnetic Resonance in Medicine (2008). This trajectory might be used in a later PR as a basis for other 3D trajectories from the literature.

A fully-sampled 2D ring trajectory consists of enough rings to respect a uniform distance of 1 / FOV between each consecutive ring. In the case where each ring is composed of exactly one shot, it reaches an acceleration factor of 2 compared to fully-sampled Cartesian. In the current implementation, any additional shot is used to further partition the ring with the longest shots, in order to reduce the maximum gradient & slew rate constraints on the trajectory.

A detailed example will be added in the near future to the other 2D trajectory examples.


paquiteau:
Nice work ! 
I think you should also add examples of the trajectory in the doc. 

PS: don't forget to use black/ruff :P

philouc:
Very nice Guillaume. Thanks for the hint!

PR #29, opened 2023-09-17
Auto register Interfaces

paquiteau:
This PR refactor the management of backend in mri-nufft. 

By auto registering interfaces (thanks to `__init_subclass__`) the backend interfaces are more easily managed. 
A small drawback: instanciation of interfaces is not directly possible (e.g. `from mrinufft import MRIfinufft` will fail)
Also This add two more conveniencies: 
- `get_operator` now accepts `*args,**kwargs` to be used for directly instanciating the NUFFT operator.
- a `list_backends` function to give a status on all backends availables. Friendlier output could also be added, for instance a `list_backends_pretty()` could  also be added, which would return something like 
```shell
>>> list_backends_pretty(verbose=True)
- finufft (available):
   requires: ["finufft"]
   < the docstring of MRIfinufft  if verbose> 
- cufinufft (not available):
   requires : ["cupy", "cufinufft"]
   < the docstring of MRICufinufft if verbose >
```
Feedback and interest on this feature would be welcome.


paquiteau:
The pretty list_backends is *"left as an exercise for the reader"* so this is LGTM ?

PR #37, opened 2023-10-02
Trajectory display overhaul

Daval-G:
This PR consists of the following features:
- New display functions to show 2D and 3D gradients with several options.
- Updates on the trajectory display functions to properly account for acquisition parameters and hardware constraints.
- Fixes on wrong units in comments and documentation.
- A new set of util functions to manipulate and convert trajectory/gradients/slew rates and check gradient constraints, as previous functions were kind of twisted to serve multiple purposes at once, notably for IO.

Currently, the documentation still needs to be written and examples updated as they are the first ones impacted by the display changes. Here is a simple rendering:

image


src/mrinufft/trajectories/display.py
```diff
@@ -57,21 +68,29 @@ def setup_3D_ticks(size, ax=None):
...
 
     Parameters
     ----------
     trajectory : array_like
         Trajectory to display.
-    size : float, optional
+    figsize : float, optional
         Size of the figure.
     one_shot : bool, optional
         If True, highlight the trajectory of the middle repetition.
```
paquiteau, line 96: Also.document the int case

src/mrinufft/trajectories/display.py
```diff
@@ -88,34 +107,51 @@ def display_2D_trajectory(
...
         ax.plot(
             trajectory[i, :, 0],
             trajectory[i, :, 1],
             color=COLOR_CYCLE[i % NB_COLORS],
             linewidth=LINEWIDTH,
         )
-    if one_shot:
+
+    # Display one shot in particular if requested
+    if one_shot or type(one_shot) == int:
```
paquiteau, line 124: Using `isinstance` is beat practice, and would check for True:

```suggestion
    if one_shot is True or insistance(one_shot, int):
```

src/mrinufft/trajectories/display.py
```diff
@@ -18,17 +24,22 @@
 ]
 NB_COLORS = len(COLOR_CYCLE)
 
-
+ALPHA = 0.2
 LINEWIDTH = 2
 POINTSIZE = 10
 FONTSIZE = 18
+SMALL_FONTSIZE = 14
```
paquiteau, line 31: Those global values are fine but I think we can do better, especially if the user wants to modify them, their change would not be taken into account in the parent module when imported. 
Something like the matplotlib `rcParams` config might be desirable.

src/mrinufft/trajectories/display.py
```diff
@@ -177,28 +219,177 @@ def display_3D_trajectory(
                 color=COLOR_CYCLE[(i + j * (not per_plane)) % NB_COLORS],
                 linewidth=LINEWIDTH,
             )
-    if one_shot:
+
+    # Display one shot in particular if requested
+    if one_shot or type(one_shot) == int:
```
paquiteau, line 224:
```suggestion
    if one_shot is True or isinstance(one_shot, int):
```

src/mrinufft/trajectories/io.py
```diff
@@ -245,13 +264,13 @@ def read_trajectory(
     grad_filename : str
         Gradient filename.
     dwell_time : float, optional
-        Dwell time of ADC, by default 0.01
+        Dwell time of ADC, by default 0.1
```
paquiteau, line 267: Missing unit

src/mrinufft/trajectories/io.py
```diff
@@ -245,13 +264,13 @@ def read_trajectory(
...
     num_adc_samples : int, optional
         Number of ADC samples, by default None
     gyromagnetic_constant : float, optional
         Gyromagnetic Constant, by default 42.576e3
     gradient_raster_time : float, optional
-        Gradient raster time, by default 0.010
+        Gradient raster time, by default 0.1
```
paquiteau, line 273: Missing unit

src/mrinufft/trajectories/utils.py
```diff
@@ -1,47 +1,156 @@
...
-DEFAULT_RESOLUTION = 6e-4  # 0.6 mm isotropic
 DEFAULT_CONE_ANGLE = np.pi / 2  # rad
 DEFAULT_HELIX_ANGLE = np.pi  # rad
 
-DEFAULT_RASTER_TIME_MS = 10e-3  # ms
-DEFAULT_GYROMAGNETIC_RATIO = 42.576e3  # MHz/T
+CARBON_GYROMAGNETIC_RATIO = 10708.4  # kHz/T
+HYDROGEN_GYROMAGNETIC_RATIO = 42576.384  # kHz/T
+PHOSPHOROUS_GYROMAGNETIC_RATIO = 17235  # kHz/T
+SODIUM_GYROMAGNETIC_RATIO = 11262  # kHz/T
```
paquiteau, line 17: I would suggest the use of a Python Enum for all those constant , and then use `Gyro.Na` or `Gyro.Sodium` (or both, Enum accepts aliases)

paquiteau (changes requested):
Seems very promising, the gradient plot are exactly what I had in mind. There is surely some docs effort to make but this is really good.

src/mrinufft/trajectories/display.py
Daval-G: It is actually one of the few failing cases of isinstance, because in Python booleans are a subclass of integers. So it would actually not work with isinstance

src/mrinufft/trajectories/display.py
Daval-G: Same here

src/mrinufft/trajectories/display.py
Daval-G: What do you mean ?

src/mrinufft/trajectories/utils.py
Daval-G: An enum has been implemented with additional values and aliases

src/mrinufft/trajectories/io.py
Daval-G: Done

src/mrinufft/trajectories/io.py
Daval-G: Done, and the value was changed back as this one was a Ctrl-H mistake.

src/mrinufft/trajectories/display.py
Daval-G: A singleton has been implemented and integrated in all display functions to enable runtime user configuration

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
+        """Initialize the display configuration with default values."""
+        self.alpha = 0.2
+        self.linewidth = 2
+        self.pointsize = 10
+        self.fontsize = 18
+        self.small_fontsize = 14
+
+
+# Initialize DisplayConfig singleton for user convenience
+displayConfig = DisplayConfig()
```
paquiteau, line 66: I think class attribute would be just as effective, and cleaner
```suggestion
class displayConfig:
    """
    A singleton used to share parameters related to display.

    Attributes
    ----------
    alpha : float
        Transparency used for area plots, by default 0.2.
    linewidth : float
        Width for lines or curves, by default 2.
    pointsize : float
        Size for points used to show constraints, by default 10.
    fontsize : float
        Font size for most labels and texts, by default 18.
    small_fontsize : float
        Font size for smaller texts, by default 14.
    """    
    alpha = 0.2
    linewidth = 2
    pointsize = 10
    fontsize = 18
    small_fontsize = 14
```

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
+        Acquisition parameters used to check on hardware constraints,
+        following the argument convention from
+        `mrinufft.trajectories.utils.compute_gradients_and_slew_rates`.
 
     Returns
     -------
     ax : plt.Axes
         Axes of the figure.
     """
+    displayConfig = DisplayConfig()
```
paquiteau, line 166: no need for it, its a global of the module.
```suggestion
```

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
     return ax
 
 
-def setup_3D_ticks(size, ax=None):
+def _setup_3D_ticks(figsize, fig=None):
     """Add ticks to 3D plot."""
-    if ax is None:
-        fig = plt.figure(figsize=(size, size))
-        ax = fig.add_subplot(projection="3d")
+    displayConfig = DisplayConfig()
```
paquiteau, line 92: No need for it, its a global of the module
```suggestion
```

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
+        `mrinufft.trajectories.utils.compute_gradients_and_slew_rates`.
 
     Returns
     -------
     ax : plt.Axes
         Axes of the figure.
     """
-    ax = setup_3D_ticks(size, subfigure)
-    trajectory = trajectory.reshape((nb_repetitions, Nc, Ns, 3))
+    displayConfig = DisplayConfig()
```
paquiteau, line 283: no need, its a global of the module
```suggestion
```

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
+    subfigure: plt.Figure or plt.SubFigure, optional
+        The figure where the trajectory should be displayed.
+        The default is `None`.
+
+    Returns
+    -------
+    axes : plt.Axes
+        Axes of the figure.
+    """
+    displayConfig = DisplayConfig()
```
paquiteau, line 403:
```suggestion
```

src/mrinufft/trajectories/display.py
```diff
@@ -19,186 +26,588 @@
...
+        Acquisition parameters used to check on hardware constraints,
+        following the argument convention from
+        `mrinufft.trajectories.utils.compute_gradients_and_slew_rates`.
+
+    Returns
+    -------
+    axes : plt.Axes
+        Axes of the figure.
+    """
+    displayConfig = DisplayConfig()
```
paquiteau, line 539:
```suggestion
```

paquiteau (changes requested):
It keeps getting better ! 
The displayConfig can be simplified though, see my suggestions

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/display.py
Daval-G: Done

paquiteau (approved):
LGTM !

PR #38, opened 2023-10-09
Add 3D wave caipi

Daval-G:
This PR adds the 3D Wave-CAIPI trajectory based on the work from Bilgic, Berkin, Borjan A. Gagoski, Stephen F. Cauley, Audrey P. Fan, Jonathan R. Polimeni, P. Ellen Grant, Lawrence L. Wald, and Kawin Setsompop. "Wave‐CAIPI for highly accelerated 3D imaging." Magnetic resonance in medicine 73, no. 6 (2015): 2152-2162. This implementation includes additional features:
- a parameter to choose the packing method, defining how the different helices should be positioned (i.e. following square or hexagonal tiling, over concentric circles or randomly)
-  a parameter to choose the shape covered over the kx-ky plane (circle, square, diamond or anything defined through p-norms)

Hereafter is a simple rendering with a square tiling of a circle, shown in 3D and over the kx-ky 2D plane.

image

image


chaithyagr:
Yay, thanks a lot for this !

Regards
Chaithya G R

philouc:
Impressive Guillaume! Many thanks

paquiteau:
Nice work ! 

I think we should also propose the canonical way of reconstructing data acquired with this trajectory (e.g FFTs and PSF deconvolution) even if its only using a dummy numpy backend, and this also open the possibility to compare it with a NUFFT based reconstruction.

Daval-G:
> Nice work !
> 
> I think we should also propose the canonical way of reconstructing data acquired with this trajectory (e.g FFTs and PSF deconvolution) even if its only using a dummy numpy backend, and this also open the possibility to compare it with a NUFFT based reconstruction.

That's a great suggestion for a future PR. I don't personally plan to work on that soon, should we create an issue to keep that in mind ?

paquiteau (changes requested):
Just a few comments... 

For the FFT I have already some code pending in pysap-fmri, that would be happy to move here (with sense support etc...). For the PSF stuff I guess we would need to find some existing implementation to adapt.

paquiteau (approved):

PR #39, opened 2023-10-13
feat: use a very large coil dimension.

paquiteau:
This PR improve the stacked nufft, by leveraging the coil dimension of the 2D operator. 
For gpu backend which does asynchronous copy of coil data (e.g. only gpuNUFFT for now, cufinufft needs a few PR being merged upstream) this is great news, and potentially leads to increase in speed.

### DIFF
--- a/src/mrinufft/operators/stacked.py
+++ b/src/mrinufft/operators/stacked.py
@@ -36,7 +36,7 @@ class MRIStackedNUFFT(FourierOperatorBase):
     # Internally the stacked NUFFT operator (self) uses a backend MRI aware NUFFT
     # operator(op), configured as such:
     # - op.smaps=None
-    # - op.n_coils = len(self.z_index) ; op.n_batchs = self.n_coils * self.n_batchs.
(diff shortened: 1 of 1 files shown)


paquiteau:
Alright some simple timing (H2D2H) reports the followings on a QUADRO P5000  with the data from the `example_stacked.py` script (181x217x181 matrix , fully sampled z axis and 8000 NUPTS in a spiral pattern) extended to have 8 coils. 

| Backend     | Forward | Adjoint |
|-------------|---------|---------|
| GPU 2D+Stack| 2.0035  | 0.643   | 
| GPU 3D      | 1.40    | 4.165   |

Overall that's a 2x speed up on the Forward + Adjoint step (= a data consistency), it can be made faster by avoiding the roundtrip in the middle, and even more when cufinufft support the async copy / computations 

The stacked 2D Nufft on GPU also has a lower memory footprint (the oversampling grid is only 4 times bigger , not 8 times) making it possible to potentially have multiple operators running together  (think fMRI)

PR #41, opened 2023-10-17
Expansion overhaul

Daval-G:
This PR is an overhaul of the former 2D-to-3D expansions, that are not 2D-to-3D only anymore. All methods are now also compatible with 3D trajectories and therefore were simply renamed to "tools". Additional tools were included, and here is a summary of all the modifications:
- stack, rotate, conify functions are extensions of the previous methods with more arguments and 3D compatibility.
- former helix expansion was removed, as it was just a garbage version of something that can already be done just with the conify function (that will be explicitly pointed out in the examples).
- two new expansion tools were added, shellify and stack_spherically. Both differ from previous expansion methods as they are using trajectory initialization functions as arguments instead of already initialized trajectories. The idea is simply that the number of shots per shell or stack will vary depending on the surface to cover, instead of creating terribly inefficient trajectories.
- the 3D FLORET trajectory was added too, as it is only a combination of initialize_2D_spiral and the conify tool.
- all previous smaller trajectory manipulation functions from trajectory3D.py were moved to tools.py.

As usual, the documentation and examples will come in later commits.

Here is an example of stack of 3D Wave-CAIPI-like helices

image


Daval-G:
I don't think the failed checks relate to the PR, can you confirm that @paquiteau ?

paquiteau:
Well, you refactored the trajectories, so now the tests importing the module fails, It would be nice of you to propagate those changes in the tests too .

PR #55, opened 2023-11-21
feat: use enums

paquiteau:
This PR makes uses of `enum` to store different key=values list, used for the trajectories. 

This makes the different options more discoverable, as well as looping over the values directly.

### DIFF
--- a/src/mrinufft/trajectories/utils.py
+++ b/src/mrinufft/trajectories/utils.py
@@ -85,9 +85,7 @@ class NormShapes(FloatEnum):
 
 
 class Tilts(str, Enum):
-    """Enumerate available tilts
-
-
+    r"""Enumerate available tilts.
 
     Notes
     -----
@@ -596,6 +594,7 @@ def initialize_tilt(tilt, nb_partitions=1):
     ------
     NotImplementedError
         If the tilt name is unknown.
+
     See Also
     --------
     Tilts
--- a/src/mrinufft/trajectories/utils.py
+++ b/src/mrinufft/trajectories/utils.py
@@ -17,6 +17,11 @@ DEFAULT_GMAX = 0.04  # T/m
 DEFAULT_SMAX = 0.1  # T/m/ms
 
 
+#########
+# ENUMS #
+#########
+
+
 class CaseInsensitiveEnumMeta(EnumMeta):
     """A case-insensitive EnumMeta."""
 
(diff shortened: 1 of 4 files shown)


src/mrinufft/trajectories/utils.py
```diff
@@ -17,7 +17,25 @@
 DEFAULT_SMAX = 0.1  # T/m/ms
```
Daval-G, line 19: Let's add an "ENUM" band to split it from regular constants

src/mrinufft/trajectories/utils.py
```diff
@@ -41,6 +59,81 @@ class Gammas(float, Enum):
     X = X129 = XENON
 
 
+class Spirals(FloatEnum):
+    """Enumerate spirals type."""
```
Daval-G, line 63: "Enumerate spiral types."

src/mrinufft/trajectories/utils.py
```diff
@@ -41,6 +59,81 @@ class Gammas(float, Enum):
...
+    ARCHIMEDES = 1
+    ARITHMETIC = ARCHIMEDES
+    FERMAT = 2
+    PARABOLIC = FERMAT
+    HYPERBOLIC = -1
+    LITHUUS = -2
+
+
+class NormShapes(FloatEnum):
+    """Enumerate norms types."""
```
Daval-G, line 74: "Enumerate shape norms."
Also I wonder if we should just name it "Shapes" to keep it similar to "Spirals", since both actually relate to values in equations

src/mrinufft/trajectories/utils.py
```diff
@@ -41,6 +59,81 @@ class Gammas(float, Enum):
...
+    INVERTED = "inverted"
+    GOLDEN = "golden"
+    MRI_GOLDEN = "mri-golden"
+    MRI = MRI_GOLDEN
+
+
+class Packings(str, Enum, metaclass=CaseInsensitiveEnumMeta):
+    """Enumerate available packing method for shots.
+
+    This is mostly use for wave-CAIPI method
```
Daval-G, line 116: "It is mostly used for the Wave-CAIPI trajectory."

Daval-G (approved):
LGTM besides a few minor details

src/mrinufft/trajectories/utils.py
Daval-G: The main part was "use**d**"

Issue #48, opened 2023-11-15
Add PROPELLER/BLADE trajectory


paquiteau:
Adding Propeller trajectories would be a nice addition to the current collection. Also this might requires some additional reconstruction methods to benefit from the motion correction.

References: 
- https://www.ajronline.org/doi/10.2214/AJR.07.3657
- https://mriquestions.com/propellerblade.html

Daval-G:
The trajectory has been added in PR #56

PR #56, opened 2023-11-21
Propeller

Daval-G:
This PR adds the Propeller/Blade trajectory (without the associated reconstruction method) as requested in #48. The original name from the authors is kept, but common aliases will be referred to in the documentation & examples coming later.

Here is an example with 6 strips of 21 shots:

image


paquiteau (approved):
LGTM so far, letme know if you want to add anything more

PR #61, opened 2023-11-23
feat: add benchmarking code

paquiteau:
This PR contains the code for the benchmark running script. 

This has still some work to do to ensure a fully reproducible pipeline (e.g. add scripts to generate the same trajectories as in ismrm, ensure the figures are the same, etc ...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@ docs/generated/
 docs_build/
 
 src/mrinufft/_version.py
+
+benchmark/outputs
+benchmark/multirun
+benchmark/trajs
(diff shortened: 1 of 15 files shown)


chaithyagr:
My personal opinion is to have this in a different repository with links in the main one. This helps decouple the main library with perf stuff making the codes more easier to maintain. Any changes in inner modules and APIs (I really hope we dont change them) will end up affecting these codes. WDYT?

Do you want to maintain them to repeat perf benchmark again to better understand how good each method is with improvements (like how we want to do that now with gpuNUFFT 2.0 vs cuFINUFFT 2.2?

chaithyagr:
What is the plan for this PR?

paquiteau:
The development of benchmark will happen in https://github.com/mind-inria/mri-nufft-benchmark

PR #66, opened 2023-12-15
Density compensation

paquiteau:
This refactors and augments the density compensation routines. 

- :heavy_check_mark: density is now a proper module, at the same level as operator and trajectories. 
- :heavy_check_mark: Added new "cell counting" method, which is purely geometric, and runs fast enough in 3D. 
- :warning: Pipe's method implementation with cufinufft has been removed (too hacky, and we want to follow the main branch of (cu)finufft). 
- :rocket: There is tests! 
- :eyes:  Density compensation methods all have the same signature `density_method(traj, shape, **kwargs)`. 

Not considered in this PR:
- Adding more density compensation method. Radio-astronomy have a lot of methods to consider and their naming can be confusing[^1][^2] 
- Registering all the density compensation methods (like we do for the operators, for now we only have 3) 
 

[^1]: https://casadocs.readthedocs.io/en/stable/notebooks/synthesis_imaging.html#Data-Weighting
[^2]: https://wsclean.readthedocs.io/en/latest/weighting.html


philouc:
Good job @paquiteau

.github/workflows/test-ci.yml
```diff
@@ -139,7 +139,7 @@ jobs:
         run: |
           cd $RUNNER_WORKSPACE
           rm -rf finufft
-          git clone https://github.com/chaithyagr/finufft --branch chaithyagr/issue306
+          git clone https://github.com/flatironinstitute/finufft
```
chaithyagr, line 142: Is there no `pip install`?

README.rst
```diff
@@ -34,7 +34,7 @@ Usage
       import numpy as np
       import mrinufft
       from mrinufft.trajectories import display
-      from mrinufft.trajectories.density import voronoi
+      from mrinufft.density import voronoi
```
chaithyagr, line 37: Do we need the voronoi as we know its like terribly slow? We can have it just there, but if no one is using it saves on maintainence

chaithyagr (changes requested):
I dont get the last point in your PR description, we technically have the same interface for density estimation?

README.rst
paquiteau: It's not slow in 2D, which is why I used it in the examples.

.github/workflows/test-ci.yml
paquiteau: As the development of cufinufft is very active, I prefer to track directly the repo.

paquiteau:
> I dont get the last point in your PR description, we technically have the same interface for density estimation?

By registering, I meant to have something to collect all density compensation methods and  then do a 
```python
get_density("cell_count")(traj, ...)  
```

.github/workflows/test-ci.yml
chaithyagr: We might also need to test against the latest release version as that's what users use.

PR #68, opened 2023-12-15
Dispatch density methods in operators

paquiteau:
This PR allow to dispatch the choice for density compensation estimation. It also provides a `get_density` methods and can show which density compensation methods are available. 

Also, It provides a generic purpose `MethodRegister` that could be used for registering other collection of methods (trajectories functions for example).

### DIFF
--- a/src/mrinufft/__init__.py
+++ b/src/mrinufft/__init__.py
@@ -35,7 +35,7 @@ from .trajectories import (
     display_3D_trajectory,
 )
 
-from .density import voronoi, cell_count, pipe
+from .density import voronoi, cell_count, pipe, get_density
 
 __all__ = [
     "get_operator",
@@ -65,6 +65,7 @@ __all__ = [
     "voronoi",
     "cell_count",
     "pipe",
+    "get_density",
 ]
 
 
--- a/src/mrinufft/_utils.py
+++ b/src/mrinufft/_utils.py
@@ -1,0 +1,82 @@
+"""General utility functions for MRI-NUFFT."""
+
+import warnings
+from collections import defaultdict
+from functools import wraps
+import numpy as np
+
+
+def proper_trajectory(trajectory, normalize="pi"):
+    """Normalize the trajectory to be used by NUFFT operators.
+
+    Parameters
+    ----------
+    trajectory: np.ndarray
+        The trajectory to normalize, it might be of shape (Nc, Ns, dim) of (Ns, dim)
+
+    normalize: str
+        if "pi" trajectory will be rescaled in [-pi, pi], if it was in [-0.5, 0.5]
+        if "unit" trajectory will be rescaled in [-0.5, 0.5] if it was not [-0.5, 0.5]
+
+    Returns
+    -------
+    new_traj: np.ndarray
+        The normalized trajectory of shape (Nc * Ns, dim) or (Ns, dim) in -pi, pi
+    """
+    # flatten to a list of point
+    try:
+        new_traj = np.asarray(trajectory).copy()
+    except Exception as e:
+        raise ValueError(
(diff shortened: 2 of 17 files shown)


src/mrinufft/__init__.py
```diff
@@ -35,7 +35,7 @@
     display_3D_trajectory,
 )
 
-from .density import voronoi, cell_count, pipe
+from .density import voronoi, cell_count, pipe, get_density
```
chaithyagr, line 38: In documentation I recommend mentioning a bit about each including computation times

src/mrinufft/_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        warnings.warn(
+            "Samples will be rescaled to [-0.5, 0.5), assuming they were in [-pi, pi)"
+        )
+        new_traj /= 2 * np.pi
+    if normalize == "unit" and np.max(new_traj) >= 0.5:
+        new_traj = (new_traj + 0.5) % 1 - 0.5
+    return new_traj
+
+
+class MethodRegister:
```
chaithyagr, line 50: Was this abstraction really needed and helpful? Especially when we have only 3 density methods and also in which only 2 are used. Although, i understand you are trying to be future proof.

src/mrinufft/operators/base.py
```diff
@@ -191,6 +152,35 @@ def with_off_resonnance_correction(self, B, C, indices):
...
+            self.density = method
+            return None
+        if not method:
+            self.density = None
+            return None
+
+        if method == "pipe" and "backend" not in kwargs:
+            method = self.pipe  # will raise error if not defined.
+        if method == "pipe" and "backend" in kwargs:
+            warnings.warn(f"Using pipe with {kwargs['backend']}")
```
chaithyagr, line 176: Why warn about it? It is more of info/log right?

src/mrinufft/operators/base.py
```diff
@@ -191,6 +152,35 @@ def with_off_resonnance_correction(self, B, C, indices):
...
+        if isinstance(method, str):
+            method = get_density(method)
+        if not callable(method):
+            raise ValueError(f"Unknown density method: {method}")
+
+        self.density = method(self.samples, self.shape, **kwargs)
```
chaithyagr, line 182: Under the current framework, how can we change shape? In case I want to compute density for a larger OSF i.e. COnventionally Pipe is done for OSF=2 (as all NUFFT is done for OSF=2)

chaithyagr: Also i am curious how can we pass additional arguments through these wrappers? I see **kwargs, but how can we use them while say instancing a NUFFT operator? Is this doable?

chaithyagr (changes requested):
Great work! Needs an example surely though to better understand the usage

src/mrinufft/operators/base.py
paquiteau: Indeed, but we are not (yet) using the logging module to provide different level of logging.

src/mrinufft/_utils.py
paquiteau: >  Although, i understand you are trying to be future proof.
Yep, Maybe we would like to collect other type of methods as well at one point (e.g. trajectories)

chaithyagr:
The call for `pipe` method must be fixed to make sure we are good with gpunufft 2,0

grid_op.op(grid_op.adj_op(density_comp, True), True) 
-->
grid_op.op(grid_op.adj_op(density_comp, None, True), None, True)

chaithyagr (approved):
Seems good to me, Great work! Maybe make a release so that I can tag in pysap-mri?

PR #76, opened 2024-01-24
feat: cherry-pick torch and lipchitz.

paquiteau:
This rebase (and squashes) the changes done in cufinufft-torch (#69) and lipchitz (#71). 

TODO: 
- [x] Add test for pytorch support
- [x] Batch test fails locally


pyproject.toml
```diff
@@ -17,7 +17,7 @@ finufft = ["finufft"]
 pynfft = ["pynfft2", "cython<3.0.0"]
 pynufft = ["pynufft"]
 
-test = ["pytest", "pytest-cov", "pytest-xdist", "pytest-sugar", "pytest-cases"]
+test = ["pytest<8.0.0", "pytest-cov", "pytest-xdist", "pytest-sugar", "pytest-cases"]
```
chaithyagr, line 20: I think its better to ensure we support >8.0.0 right? Is it a big issue, maybe open one?

src/mrinufft/_utils.py
```diff
@@ -4,6 +4,57 @@
 from collections import defaultdict
 from functools import wraps
 import numpy as np
+from numpy.typing import DTypeLike
```
chaithyagr, line 7: Maybe have a look at https://github.com/pearu/pydlpack
This is pretty much what you want to do right?

src/mrinufft/density/utils.py
```diff
@@ -1,4 +1,5 @@
 """Utilities for density compensation."""
+
```
chaithyagr, line 2: Lets not add unnecessary spaces in history to mess things up? This can be a separate PR if it is docs or codestyle related right?

tests/helpers/factories.py
```diff
@@ -24,3 +39,61 @@ def kspace_from_op(operator):
         operator.cpx_dtype
     )
     return kspace
+
+
+def to_interface(data, interface):
```
chaithyagr, line 44: SHouldnt this be avvailable to users ? I know it happens internally, but in case user wants to do something also?

chaithyagr (changes requested):
Good work!, this is just first round of review. I might strongly recommend using dlpack if possible to prevent unnecessary copies in future if the working of functions change.

pyproject.toml
paquiteau: Well https://github.com/smarie/python-pytest-cases/issues/330 is already there. 

I don't know (yet) how to remove the need for pytest-cases, we rely heavily on it for the parametrization...

tests/helpers/factories.py
paquiteau: This mostly for helping the test setup. Inside MRI-nufft we have the `@with_numpy` decorator for all the backend working internally with numpy. For now, only cufinufft has some custom handling (for pytorch and cupy). 

Indeed such function are usefull, but IMO they are not central to the use of MRI-nufft (the user comes with its array, we take care of the rest). Maybe we could add them to more generic library like Modopt.

src/mrinufft/density/utils.py
paquiteau: Let me try some git wizardry to handle that.

src/mrinufft/_utils.py
paquiteau: welll the typing annotation is just fancy for now. 

From what I understood, all the array libraries have a `from_dlpack` so we don't need the python package you mention earlier. And we still need to handles the device <-> host communication 
```
>>> import numpy as np
>>> import cupy as cp 
>>> np.from_dlpack(cp.arange(10))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Unsupported device in DLTensor.
```

chaithyagr:
LGTM, please proceed to merge when green

PR #78, opened 2024-02-02
fix: robust voronoi estimation.

paquiteau:
Voronoi density compensation may have some trouble estimating area/volumes of cell that are on boundary, this PR removes the outliers and uses interpolation to fix values.


paquiteau:
image

philouc (approved):
This is what we call a good bug fixing!

PR #82, opened 2024-02-21
feat: start an autodiff poc.

paquiteau:


chaithyagr:
Fixed the tests here ! Good work guys, although we need to merge #96 first

tests/test_autodiff.py
```diff
@@ -0,0 +1,103 @@
...
+)
+def operator(kspace_loc, shape, backend, autograd):
+    """Create NUFFT operator with autodiff capabilities."""
+    kspace_loc = kspace_loc.astype(np.float32)
+
+    nufft = get_operator(backend_name=backend, autograd=autograd)(
+        samples=kspace_loc,
+        shape=shape,
+        smaps=None,
+        squeeze_dims=False,  # Squeezing breaks dimensions !
```
chaithyagr, line 43: @paquiteau I think we need a check for this. We need squeeze_dims to be false for now if we want autograd to work good.

chaithyagr: Added check for this.

src/mrinufft/operators/autodiff.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Parameters
+    ----------
+    nufft_op: Classic Non differentiable MRI-NUFFT operator.
+    """
+
+    def __init__(self, nufft_op):
+        super().__init__()
+        if nufft_op.squeeze_dims:
+            raise ValueError("Squeezing dimensions is not "
```
chaithyagr, line 52: Here, we check if squeeze_dims is True, we raise an error

PR #85, opened 2024-03-06
A major bug in NSP read trajectory implementation

chaithyagr:
This bug has wide implications. Bascially when OSF>1 (which is usually the case), for the first 0, 1, ... OSF points for each shot, we end up with bad k-space trajectory location which is wrong!


chaithyagr:
Noting here that the issue mostly affects center-out trajectories adversly

paquiteau:
good catch ! you might want to lint your code however ;)

chaithyagr:
I just thought i did xD didnt commit damn it.

chaithyagr:
This is bizzare, I thought this was merged! @paquiteau any idea how we missed this, can I merge it?

Issue #72, opened 2023-12-19
Trajectory issue: Min value out of [-0.5,0.5] range


dennishnf:
After creating 3D cones trajectories, I have observed that he min value of the trajectory if out f the [-0.5,0.5] range. Here I attached the code to reproduce this issue:

```
import mrinufft

# Trajectory parameters
Nc = int(208*384/10)  # Number of shots
Ns = 384 # Number of samples per shot
in_out = False  # Choose between in-out or center-out trajectories
tilt = "uniform"  # Angular distance between shots

trajectory = mrinufft.initialize_3D_cones(Nc, Ns, in_out=in_out)

trajectory = trajectory.reshape(Nc*Ns,3)

# here min value out of [-0.5,0.5] range
trajectory.min()
-0.5000150076299451
```

chaithyagr:
I think some of the trajectories will slightly exceed the constraints just to ensure we still satisfy the hardware constraints. I will let @Daval-G comment if this is a real bug to be fixed. Else, we can close this.

paquiteau:
Addditionally we might want to partially change the behaviour of [`proper_trajectory`](https://github.com/mind-inria/mri-nufft/blob/eda0e6298adf2b7f2c6efbd3e9fe208d978528f3/src/mrinufft/_utils.py#L9):
- currently if the trajectory lies in a range extended by epsilon [^1] the data is not rescale and used as it. 
- We could either truncate (or wrap ?) coordinates that are outside the range, but inside the epsilon region 

[^1]: epsilon = 1e-4, it's hardcoded for now.

Daval-G:
This problem comes from a quite approximative initialization of the cones specific to 3D Cones, and particularly the way their width were computed. It should be fixed in PR #86 because the 3D Cones trajectory now relies on our own tools and 2D spirals, along with a calculation of the cone widths based on optimal surface sphere packing.

Daval-G:
Based on the example provided, it seems that the problem is fixed. Can you confirm it's OK on your side @dennishnf ?

Quick note: I don't recommend using uniform tilt for 3D Cones, the default one is golden tilt because it allows the cones to match a Fibonacci lattice of the 3D k-space sphere, which is approximately well-balanced.

PR #86, opened 2024-04-03
Fibonacci patch

Daval-G:
This PR patches the bug from issue #72 and changes a few features. Most of them are related to the use of Fibonacci lattices, hence the name:
- The `precess` tool was previously introduced mainly to provide a way to rotate shots such as 3D Cones and Seiffert spirals following the [Fibonacci sphere lattice approximation ](https://observablehq.com/@meetamit/fibonacci-lattices) to provide a mostly uniform distribution of shots over the k-space sphere. However it was subtly wrong in the way polar angles were handled, as shown below with the clouds of points used to rotate single shots. It resulted in a greater density around the precession axis instead:
Previously wrong method:

New correct method:

The reason was simply that the kz coordinate was handled in a polar fashion rather than an axial fashion, as shown below. Now both options are available, and the `precess` tool is designed to provide a proper Fibonacci sphere distribution with its default arguments. The examples were also largely extended to better explain this tool.
Previously wrong method (polar partition):

New correct method (axial partition):

- All rotation and prime number-related functions used in trajectories were moved to a new `maths.py` file. The `Rv` function that returns a rotation matrix based on two vectors using Rodrigues' rotation coefficients now handles properly co-linear cases, and a new function has been added to provide rotation matrices around arbitrary vectors (used in Seiffert spirals).
- Functions are now provided in `maths.py` to generate Fibonacci square, circle and sphere lattices. Those are used to approximate analytically a uniform distribution of shot positions/orientations, typically over a sphere surface.
- 3D Cones trajectory is now based on initialize_2D_spiral (which extend considerably its customization), but also the `conify` tool, which solves the problem signaled in issue #72, and the fixed `precess` tool with proper Fibonacci sphere lattice.
- Wave-CAIPI trajectory has a new "fibonacci" packing option based on Fibonacci circle lattice.
- Seiffert spirals are more closely following the original paper, with spiral precessions based on their own axes using the newly fixed `precess` tool with proper Fibonacci sphere lattice, but also around their own axes to avoid azimuthal redundancy, following authors methodology.
- FLORET trajectory initialization has been changed to better match the definition from its original paper and avoid confusion on its nature compared to similar trajectories also available (conic spirals/stack of cones, for example).


src/mrinufft/trajectories/utils.py
```diff
@@ -1,5 +1,6 @@
 """Utility functions for the trajectory design."""
 
+import numbers as nb
```
paquiteau, line 3: `nb` is often used for numba. I would rather import the numbers type you want directly. 

```suggestion
from numbers import Real
```

src/mrinufft/trajectories/utils.py
```diff
@@ -630,7 +468,7 @@ def initialize_tilt(tilt, nb_partitions=1):
     Tilts
 
     """
-    if not (isinstance(tilt, str) or tilt is None):
+    if isinstance(tilt, nb.Number):
```
paquiteau, line 471:
```suggestion
    if isinstance(tilt, Real):
```

src/mrinufft/trajectories/utils.py
```diff
@@ -661,7 +499,7 @@ def initialize_spiral(spiral):
     float
         Spiral power value.
     """
-    if isinstance(spiral, float):
+    if isinstance(spiral, nb.Number):
```
paquiteau, line 502:
```suggestion
    if isinstance(tilt, Real):
```

src/mrinufft/trajectories/utils.py
```diff
@@ -679,6 +517,6 @@ def initialize_shape_norm(shape):
     float
         Shape p-norm value.
     """
-    if isinstance(shape, float):
+    if isinstance(shape, nb.Number):
```
paquiteau, line 520:
```suggestion
    if isinstance(tilt, Real):
```

src/mrinufft/trajectories/maths.py
```diff
@@ -0,0 +1,313 @@
...
+        Second integer.
+
+    Returns
+    -------
+    int
+        The greatest common divider of p and q.
+    """
+    while q != 0:
+        p, q = q, p % q
+    return p
```
paquiteau, line 32: I guess this could be replaced with [`np.gcd`](https://numpy.org/doc/stable/reference/generated/numpy.gcd.html) ?

paquiteau:
Awesome stuff ! I have made small suggestions, but overall this is good for me :)

philouc:
Congrats @Daval-G for this awesome upgrade on handling 3D Cone trajectories!

src/mrinufft/trajectories/utils.py
Daval-G: Done

src/mrinufft/trajectories/utils.py
Daval-G: Done (without suggestion, because `spiral` was replaced by `tilt`)

src/mrinufft/trajectories/utils.py
Daval-G: Done (without suggestion, because `shape` was replaced by `tilt`)

src/mrinufft/trajectories/maths.py
Daval-G: Indeed, the function is removed and its single use has been replaced & checked

PR #87, opened 2024-04-04
share a 2D nufft across Stacked Operators

paquiteau:
This PR allows one to initialize a 2D NUFFT operators and reuse it across multiple definition of stacked (2.5D) NUFFT. 

Tests are coming.

### DIFF
--- a/src/mrinufft/operators/stacked.py
+++ b/src/mrinufft/operators/stacked.py
@@ -6,7 +6,12 @@ import numpy as np
 import scipy as sp
 
 from mrinufft._utils import proper_trajectory, power_method, get_array_module, auto_cast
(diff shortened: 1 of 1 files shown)


chaithyagr:
There seems to be conflicts... Can you please handle that. Also, can you please give the usecase for this? Is it for the specific case when you shall have different number of stacks so you want to reuse the 2D operator, or is it when you wangt to reconstruct different time frame volumes for a given stack ? Is there any expected performance gain (in terms of creating and destroying the backend operator)

paquiteau:
the typical use case is for this kind of trajectory, in the context of fMRI: 
image

The 2D shot is the same everywhere, so instead of creating a new NUFFT plan for each frame, we can reuse it. This saves a significant amount of memory.

PR #89, opened 2024-04-10
Added siemens data reader and phase shifter

chaithyagr:
These essential functions seem to be better if it belonged here as compared to on `pysap-mri`. That way, you can have a complete DC Adjoint code working from CLI without any need for pysap-mri, for light simple quick rersults..

Later, `pysap-mri` can build further ahead on this CLI for compressed sensing based recon. I have basic CLI code setup and I shall make another PR for it. Still figuring out the structuring of config.


src/mrinufft/io/siemens.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+
+try:
+    from mapvbvd import mapVBVD
```
paquiteau, line 4: This needs to also be added to `pyproject.toml` (maybe in an `io` dependency bundle ?)

src/mrinufft/io/siemens.py
```diff
@@ -0,0 +1,87 @@
...
+    -----
+    This function requires the mapVBVD module to be installed. 
+    You can install it using the following command:
+        `pip install pymapVBVD`
+    """
+    if not MAPVBVD_FOUND:
+        raise ImportError(
+            "The mapVBVD module is not available. Please install it using "
+            "the following command: pip install pymapVBVD"
+        )
```
paquiteau, line 45: you may simply do the import in the function then ?

src/mrinufft/io/siemens.py
```diff
@@ -0,0 +1,87 @@
...
+        Whether to remove the oversampling, by default False.
+    squeeze : bool, optional
+        Whether to squeeze the dimensions of the data, by default True.
+    data_type : str, optional
+        The type of data to read, by default 'SPARKLING_VE11C'.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the raw data and the header information.
```
paquiteau, line 28: describe what the tuple does 
```suggestion
    data: ndarray
        Imported data formatted as XXX 
    hdr: dict
        Extra information about the data parsed from the twix file
```

src/mrinufft/io/siemens.py
```diff
@@ -0,0 +1,87 @@
...
+    twixObj.image.flagRemoveOS = removeOS
+    twixObj.image.squeeze = squeeze
+    raw_kspace = twixObj.image['']
+    data = np.moveaxis(raw_kspace, 0, 2)
+    hdr = {
+        "num_coils": int(twixObj.image.NCha),
+        "num_shots": int(twixObj.image.NLin),
+        "num_contrasts": int(twixObj.image.NSet),
+        "num_adc_samples": int(twixObj.image.NCol),
+        "num_slices": int(twixObj.image.NSli),
```
paquiteau, line 58: overall in MRI-NUFFT we tend to use `n_coils`, `n_shots`, etc... I would advise to use these convention here too.

src/mrinufft/io/siemens.py
```diff
@@ -0,0 +1,87 @@
...
+    }
+    data = data.reshape(
+        hdr["num_coils"], 
+        hdr["num_shots"]*hdr["num_adc_samples"], 
+        hdr["num_slices"], 
+        hdr["num_contrasts"]
+    )
+    if "SPARKLING_VE11C" in data_type:
```
paquiteau, line 66: do you have other value of data_type in mind ? 
I would recommend to use an enum to gather them. Overall this section of the code only updates the header, and may be refactored to a dedicated function. The workflow would then be 

```python 
parser = parse_header_type[data_type] # e.g. using a `MethodRegister` from `_utils.py`
hdr |= parser(twixObj)
```

src/mrinufft/io/utils.py
```diff
@@ -0,0 +1,38 @@
+from mrinufft.operators.base import with_numpy_cupy, get_array_module
+
+
+@with_numpy_cupy
+def add_phase_to_kspace_with_shifts(kspace_data, kspace_loc, normalized_shifts):
```
paquiteau, line 5: `with_numpy_cupy` expect a signature with (self, data, output=None). I think you are going to get trouble here.

paquiteau (changes requested):
Nice proposal. There are some points to discuss though. I think that mri-nufft should remains as agnostic and generalist as possible. I view it more as a backend stuff, on which people build their stuff (while citing us, of course :D )

chaithyagr:
Adding back for review of updated codes. I will get some tests up also though

src/mrinufft/io/nsp.py
```diff
@@ -253,7 +253,7 @@ def read_trajectory(
     grad_filename: str,
     dwell_time: float = DEFAULT_RASTER_TIME,
     num_adc_samples: int = None,
-    gamma: Union[Gammas,float] = Gammas.HYDROGEN,
+    gamma: Union[Gammas, float] = Gammas.HYDROGEN,
```
paquiteau, line 256:
```suggestion 
    gamma: Gammas | float = Gammas.HYDROGEN,
```

paquiteau (changes requested):

src/mrinufft/io/nsp.py
```diff
@@ -390,3 +390,89 @@ def read_trajectory(
...
+        hdr["n_contrasts"],
+    )
+    if "ARBGRAD_VE11C" in data_type:
+        hdr["type"] = "ARBGRAD_GRE"
+        hdr["shifts"] = ()
+        for s in [7, 6, 8]:
+            shift = twixObj.search_header_for_val(
+                "Phoenix", ("sWiPMemBlock", "adFree", str(s))
+            )
+            hdr["shifts"] += (0,) if shift == [] else (shift[0],)
```
paquiteau, line 466: don't you want to apply the shift as well here ?

src/mrinufft/io/nsp.py
```diff
@@ -390,3 +390,89 @@ def read_trajectory(
...
+            "Phoenix", ("sWiPMemBlock", "alFree", "4")
+        )[0]
+        hdr["trajectory_name"] = twixObj.search_header_for_val(
+            "Phoenix", ("sWipMemBlock", "tFree")
+        )[0][1:-1]
+        if hdr["n_contrasts"] > 1:
+            hdr["turboFactor"] = twixObj.search_header_for_val(
+                "Phoenix", ("sFastImaging", "lTurboFactor")
+            )[0]
+            hdr["type"] = "ARBGRAD_MP2RAGE"
```
paquiteau, line 477: This seems to be highly specific to your usecase for MP2RAGE reconstruction, are you sure you want this to be here ?

paquiteau (commented):
Nice work. I still think we should extract the ARBGRAD stuff in a dedicated function.

This could be a wrapper around `read_siemens_rawdat`, e.g a `read_nsp_mp2rage` or alike.

src/mrinufft/io/nsp.py
chaithyagr: Nope :) I think thats very specific to processing and not just basic IO

chaithyagr:
There seems to be no way to test IO arbGrad module. For now lets just merge and add an issue for needing test?

chaithyagr:
Something went wrong here! I dont see my changes at master. The split up of read_siemens_rawdat and read_arbgrad_rawdat stuff. I will change this in #90 PR.

PR #96, opened 2024-04-26
Fix NDFT tests

chaithyagr:
The fixes include:
1) fftshift added for FFT <-> NDFT matrix test
2) Added a test to compare NDFT vs NUFFT. This is crucial and sets the tone for autodiff modules.
3) Adding normalize to support `ortho`, this is important as mostly all NUFFT is normalized. 
4) Removed a test `test_ndft_grid_matrix`, which tries o get the FFT matrix and match to a gridded NDFT matrix. But this is too complex and not generalizable currently. The test seems redundant so I removed it.


tests/test_ndft.py
```diff
@@ -13,32 +13,9 @@
 
 from case_trajectories import CasesTrajectories, case_grid1D
 from helpers import assert_almost_allclose
+from mrinufft import get_operator
 
 
-@parametrize_with_cases(
```
chaithyagr, line 18: This test is bad. Removed it :P

tests/test_ndft.py
```diff
@@ -76,7 +53,32 @@ def test_ndft_implicit1(kspace, shape):
     linop_coef = implicit_type1_ndft(kspace, random_kspace.flatten(), shape)
     matrix_coef = matrix.conj().T @ random_kspace.flatten()
 
-    assert np.allclose(linop_coef, matrix_coef)
+    assert_almost_allclose(linop_coef, matrix_coef, atol=1e-4, rtol=1e-4, mismatch=5)
```
chaithyagr, line 56: This test is good, so added it  :P

paquiteau:
Good catch ! could you detail a bit which things were wrong (and why) ?
Also, lint your code ;)

chaithyagr:
@paquiteau Ill let you add this test for CD! I think that its good practice to run it before every release. (More importantly, we need to make it run right now to make sure the tests run fine.)

chaithyagr:
Can i merge when green? @paquiteau

PR #97, opened 2024-04-26
Gpunufft data consistency

paquiteau:
This PR add support for on-gpu data_consistency operation with GpuNUFFT. 

There are still some corner tests that fails on my machine locally.

### DIFF
--- a/src/mrinufft/operators/base.py
+++ b/src/mrinufft/operators/base.py
@@ -140,13 +140,13 @@ def with_numpy_cupy(fun):
             output_ = output
 
         if output_ is not None:
-            if not((
-                is_host_array(data_) and is_host_array(output_)
-            ) or (
-                is_cuda_array(data_) and is_cuda_array(output_))
+            if not (
+                (is_host_array(data_) and is_host_array(output_))
+                or (is_cuda_array(data_) and is_cuda_array(output_))
             ):
-                raise ValueError("input data and output should be "
-                                 "on the same memory space.")
+                raise ValueError(
+                    "input data and output should be " "on the same memory space."
+                )
         ret_ = fun(self, data_, output_, *args, **kwargs)
 
         if xp.__name__ == "torch" and is_cuda_array(data):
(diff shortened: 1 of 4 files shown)


src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -542,3 +545,103 @@ def _safe_squeeze(self, arr):
...
+
+        # dispatch
+        if is_host_array(image_data) and is_host_array(obs_data):
+            grad_func = self._dc_host
+        elif is_cuda_array(image_data) and is_cuda_array(obs_data):
+            grad_func = self._dc_direct
+        else:
+            raise ValueError("image and obs_data should be on the same device")
+
+        if not self.impl.use_gpu_direct:
```
chaithyagr, line 598: To me this warning doesnt make sense here. This is should be reported only if the input is a GPU array

src/mrinufft/operators/base.py
```diff
@@ -139,6 +139,13 @@ def wrapper(self, data, output=None, *args, **kwargs):
             data_ = data
             output_ = output
 
+        if output_ is not None:
+            if not((
+                is_host_array(data_) and is_host_array(output_)
+            ) or (
+                is_cuda_array(data_) and is_cuda_array(output_))
+            ):
```
chaithyagr, line 147: I added these lines here. I think it is a good check. For now we dont support CPU / GPU input outputs

src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -314,7 +316,7 @@ def adj_op_direct(self, coeffs, image=None, grid_data=False):
                 order="F",
             )
         self.operator.adj_op_direct(coeffs.data.ptr, image.data.ptr, grid_data)
-        image = image.reshape(C, *self.shape[::-1])
+        image = image.T.reshape(C, *self.shape[::-1], order="C")
```
chaithyagr, line 319: This seems to have been the issue.
`<>.T.reshape(..., order="C")` is not the same as `<>.reshape(..., order="F")`

PR #105, opened 2024-05-19
Spiral patch

Daval-G:
This PR updates or adds the following features, summarized graphically in the picture down below:
1. **Update spiral parameterization from radial to polar**: Algebraic spirals (also known as [general archimedean spirals](https://en.wikipedia.org/wiki/Archimedean_spiral#General_Archimedean_spiral)) are usually defined in polar coordinate with the radius as a function of the polar angle (polar parameterization), but the opposite can be done too. It was a tricky question that was solved previously by defining the polar angle as a function of the radius (radial parameterization) because the implementation was convenient. However it makes the radius increase by a constant value, which leads to non-senses when referring to the literature, 3D FLORET in particular. Indeed, the goal of using Fermat spirals is to reduce the amount of points close to the center, increase them at the edges to achieve less central redundancy and a better edge coverage while decreasing the gradient strength required. The opposite happens with radial parameterization, and therefore it was changed to polar parameterization. This change mostly improve Fermat-based trajectories, while Archimedes spirals are not affected at all by this change. The power values were reversed to match this polar parameterization. However polar parameterization creates anomalies at the center for spirals with powers superior to 2 such as Fermat, fixed with a new feature explained below.
2. **Remove lithuus, hyperbolic and all spirals with negative powers**: Algebraic spirals with negative power values such as Lithuus or Hyperbolic spirals are irrelevant for MRI as their behavior is asymptotically chaotic when approaching the center, without possibility to fix it. The `initialize_2D_spiral` function now raises an error when negative values are given as arguments.
3. **Add 2D Fibonacci spiral trajectory**: Some literature also exists about non-algebraic spirals, and I proposed to reproduce the one from _Cline, Harvey E., and Thomas R. Anthony. "Uniform k-space sampling with an interleaved Fibonacci spiral acquisition." In Proceedings of the 7th Annual Meeting of ISMRM, Philadelphia, USA, vol. 1657. 1999._  as `initialize_2D_fibonacci_spiral`. The idea is to produce a strictly uniform density by using Fibonacci circle lattices, split into individual center-out spirals that are slightly eccentric to avoid redundancy. The number of spirals is required to be on the Fibonacci sequence, and therefore additional helper functions are provided in `mrinufft.trajectories.maths`.
4. **Add function to fix gradient central anomalies**: Most spiral-based trajectories from the literature have slew rate anomalies close to the center or poles (algebraic spirals with polar parameterization, Fibonacci spirals, Seiffert spirals, Concentric/Helical shells). A function is added to `mrinufft.trajectories.gradients` to enforce non-strictly increasing monoticity. This fix is added as an option to `initialize_2D_spirals` and `initialize_2D_fibonacci_spiral`, and will be added soon to `initialize_3D_seiffert_spiral` and `initialize_3D_helical_shell` in another PR.
5. **Add alias for Galilean spirals**: An alias is added for Galilean spirals, another type of algebraic spirals.

This PR is the first one to propose a function to adjust gradients, but more are coming as it is a major element of trajectory design, often present in the literature.

Screenshot from 2024-05-19 15-29-01

### DIFF
--- a/src/mrinufft/trajectories/gradients.py
+++ b/src/mrinufft/trajectories/gradients.py
@@ -1,0 +1,40 @@
+import numpy as np
+import numpy.linalg as nl
+from scipy.interpolate import CubicSpline
+
+
+def patch_center_monoticity(parameters, update_shot, update_parameters, in_out=False, learning_rate=1e-1):
+    single_shot = update_shot(*parameters)
+    Ns = single_shot.shape[0]
+
+    old_index = 0
+    old_x_axis = np.zeros(Ns)
+    x_axis = np.linspace(0, 1, Ns)
+
+    while nl.norm(x_axis - old_x_axis) / Ns > 1e-10:
+        # Determine interval to fix
+        single_shot = update_shot(*parameters)
+        gradient_norm = nl.norm(np.diff(single_shot[(Ns // 2) * in_out:], axis=0), axis=-1)
+        arc_length = np.cumsum(gradient_norm)
+
+        index = gradient_norm[1:] - arc_length[1:] / (1 + np.arange(1, len(gradient_norm)))
+        index = np.where(index < 0, np.inf, index)
+        index = max(old_index, 2 + np.argmin(index))
+        start_index = (Ns // 2 + (Ns % 2) - index) * in_out
+        final_index = (Ns // 2) * in_out + index
+
+        # Balance over interval
+        cbs = CubicSpline(x_axis, single_shot.T, axis=1)
+        gradient_norm = nl.norm(np.diff(single_shot[start_index:final_index], axis=0), axis=-1)
+
+        old_x_axis = np.copy(x_axis)
+        new_x_axis = np.cumsum(np.diff(x_axis[start_index:final_index]) / gradient_norm)
+        new_x_axis *= (x_axis[final_index - 1] - x_axis[start_index]) / new_x_axis[-1]  # Normalize
+        new_x_axis += x_axis[start_index]
+        x_axis[start_index + 1:final_index - 1] += (new_x_axis[:-1] - x_axis[start_index + 1:final_index - 1]) * learning_rate
+
+        # Update loop variables
+        old_index = index
+        single_shot = cbs(x_axis).T
+        parameters = update_parameters(single_shot, *parameters)
+    return parameters
--- a/src/mrinufft/trajectories/gradients.py
+++ b/src/mrinufft/trajectories/gradients.py
@@ -1,23 +1,93 @@
+"""Functions to improve/modify gradients."""
+
 import numpy as np
 import numpy.linalg as nl
 from scipy.interpolate import CubicSpline
 
 
-def patch_center_monoticity(parameters, update_shot, update_parameters, in_out=False, learning_rate=1e-1):
-    single_shot = update_shot(*parameters)
+def patch_center_anomaly(
+    shot_or_params,
+    update_shot=None,
+    update_parameters=None,
+    in_out=False,
+    learning_rate=1e-1,
+):
+    """Re-position samples to avoid center anomalies.
+
+    Some trajectories behave slightly differently from expected when
+    approaching definition bounds, most often the k-space center as
+    for spirals in some cases.
+
+    This function enforces non-strictly increasing monoticity of
+    sample distances from the center, effectively reducing slew
+    rates and smoothing gradient transitions locally.
+
+    Shots can be updated with provided functions to keep fitting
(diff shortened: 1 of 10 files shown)


src/mrinufft/trajectories/gradients.py
```diff
@@ -0,0 +1,118 @@
+"""Functions to improve/modify gradients."""
+
+import numpy as np
+import numpy.linalg as nl
+from scipy.interpolate import CubicSpline
+
+
+def patch_center_anomaly(
```
paquiteau, line 8: This is a really good addition to MRI-nufft !

src/mrinufft/trajectories/maths.py
```diff
@@ -210,6 +210,55 @@ def Ra(vector, theta):
 #############
 
 
+def is_from_fibonacci_sequence(n):
+    """Check if an integer belongs to the Fibonacci sequence.
+
+    An integer belongs to the Fibonacci sequence if either
+    :math:`5*n²+4` or :math:`5*n²-4` is a perfect square.
```
paquiteau, line 217: maybe add a reference (ex: https://en.wikipedia.org/wiki/Fibonacci_sequence#Recognizing_Fibonacci_numbers)

src/mrinufft/trajectories/maths.py
```diff
@@ -210,6 +210,55 @@ def Ra(vector, theta):
...
+    ----------
+    n : int
+        Integer to match.
+
+    Returns
+    -------
+    int
+        Closest number from the Fibonacci sequence.
+    """
+    # Generate Fibonacci number up to n
```
paquiteau, line 250: I think you can do better by using F_n ≈ int(phi^n/sqrt(5)) and thus `floor(log_phi(x))` should give you the closest index, from which you can compute F_n. 

This would probably be more performant for large n, but I don't think we care for that here.

src/mrinufft/trajectories/trajectory2D.py
```diff
@@ -64,27 +80,131 @@ def initialize_2D_spiral(
...
     array_like
         2D spiral trajectory
     """
-    # Initialize a first shot in polar coordinates
-    segment = np.linspace(-1 if (in_out) else 0, 1, Ns)
-    radius = KMAX * segment
-    angles = 2 * np.pi * nb_revolutions * (np.abs(segment) ** initialize_spiral(spiral))
+    # Negative spirals like hyperbolic or lithuus have
+    # asymptotic behavior around the center, making them
+    # irrelevant for MRI or real life applications.
```
paquiteau, line 95: rather add this to a Notes section in the docstring ?

paquiteau (changes requested):
Awesome work ! Nothing much to change, I just made a few comments for the posterity. Good to merge once the CI is happy.

src/mrinufft/trajectories/maths.py
Daval-G: The check between upper and lower Fibonacci number still seems to be necessary, but getting them is now a constant time computation

src/mrinufft/trajectories/maths.py
Daval-G: Done

Daval-G:
I had to update the gif examples to make them work again, and I took the occasion to clean them a bit

paquiteau:
I think you cleaned up the gif examples a bit too much. the docs fails to build locally. (try `python -m docs docs_build` to check it)

```
../examples/example_gif_3D.py unexpectedly failed to execute correctly:

    Traceback (most recent call last):
      File "/volatile/home/pc266769/gits/nsp/code/mri-nufft/examples/example_gif_3D.py", line 203, in <module>
        final_dir = Path(__file__).parent / "docs" / "generated" / "autoexamples" / "images"
    NameError: name '__file__' is not defined

WARNING: g gallery for generated/autoexamples... [ 73%] example_gif_2D.py
../examples/example_gif_2D.py unexpectedly failed to execute correctly:

    Traceback (most recent call last):
      File "/volatile/home/pc266769/gits/nsp/code/mri-nufft/examples/example_gif_2D.py", line 198, in <module>
        final_dir = Path(__file__).parent / "docs" / "generated" / "autoexamples" / "images"
    NameError: name '__file__' is not defined
```
I must admit that don't know why this happens, since you apparently did not touch the file generation parts of the script.

Daval-G:
The gifs are not appearing in the current master documentation either, is it expected ?

paquiteau:
okay, in this case it will be tackle separately I guess (typically in the trajectory gallery PR we are avoiding for now)

Daval-G:
Good to merge ? @paquiteau

PR #111, opened 2024-05-22
Add torch kbnufft support

Lenoush:
add torchkbnufft.py (https://github.com/mind-inria/mri-nufft/blob/101-add-torch-kbnufft/src/mrinufft/operators/interfaces/torchkbnufft.py ) based on tfnufft.py (https://github.com/mind-inria/mri-nufft/blob/101-add-torch-kbnufft/src/mrinufft/operators/interfaces/tfnufft.py)

librairy use : https://torchkbnufft.readthedocs.io/en/stable/index.html
where estimate_density become calc_density_compensation_function
and we use KbNufft and KbNufftAdjoint for applying NUFFT operations

im_size have to be a tuple of the size of the base image so it's shape[2:] without  coils and batchs
omega represents the Kspace trajectory


src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,115 @@
...
+
+        kb_ob = torchmri.KbNufft(
+            im_size=self.shape[2:],
+        )
+        return kb_ob(image=data_d, omega=self.samples, smaps=self.smaps) 
+
+    def adj_op(self, data):
+        """
+        Backward Operation.
+``image`` calculated at scattered Fourier locations.
```
chaithyagr, line 95: Please add tab here

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,115 @@
...
+        Parameters
+        ----------
+        data: Tensor
+
+        Returns
+        -------
+        Tensor
+        """
+        if self.uses_sense:
+            data_d = data * self.smaps
```
chaithyagr, line 83: You dont need this, the underlying KBNUFFT uses Smaps already. Please provide the Smaps to the NUFFT module so that calculations are right

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,115 @@
...
+    density: bool or Tensor
+       Density compensation support.
+        - If a Tensor, it will be used for the density.
+        - If True, the density compensation will be automatically estimated,
+          using the fixed point method.
+        - If False, density compensation will not be used.
+    smaps: Tensor
+    """
+
+    backend = "torch"
```
chaithyagr, line 37: Please add this as `torchkbnufft`

chaithyagr (changes requested):

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,115 @@
...
+        -------
+        Tensor
+        """
+        if self.uses_sense:
+            data_d = data * self.smaps
+        else:
+            data_d = data
+
+        kb_ob = torchmri.KbNufft(
+            im_size=self.shape[2:],
```
paquiteau, line 88: This plan can be initialized in the init method and saved as a private attribute, to avoid to initialize a new operator at each call. 

```
# in init
self._tkb_op = torchmri.KbNufft 


# here 
return self._tkb_op(....)
```

Same for the adjoint operator below.

src/mrinufft/operators/interfaces/torchkbnufft.py
Lenoush: I don't understand you want me to put smaps parameter in KbNufft ?
I can't because the class isn't defined that way: https://torchkbnufft.readthedocs.io/en/stable/generated/torchkbnufft.KbNufft.html

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,110 @@
+"""Pytorch MRI Nufft Operators."""
+
+from ..base import FourierOperatorBase
+
+TORCH_AVAILABLE = True
+
+try:
+    import torchkbnufft as torchmri
```
chaithyagr, line 8: maybe call it torchnufft?

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,110 @@
...
+    available = TORCH_AVAILABLE
+
+    def __init__(self, samples, shape, n_coils=1, density=False, smaps=None, eps=1e-6):
+        super().__init__()
+
+        self.samples = samples
+        self.shape = shape
+        self.n_coils = n_coils
+        self.eps = eps
+        self._tkb_op = torchmri.KbNufft
```
chaithyagr, line 47: You need to initialize the class here.
Same for adjoint

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,110 @@
...
+        Parameters
+        ----------
+        data: Tensor
+
+        Returns
+        -------
+        Tensor
+        """
+        kb_ob = self._tkb_op(im_size=self.shape[2:])
+        return kb_ob(image=data, omega=self.samples, smaps=self.smaps)
```
chaithyagr, line 85: I might recommend against doing this, why do it again?

chaithyagr: Like why initialize every time? @paquiteau , would this be best comparision?

chaithyagr (changes requested):

src/mrinufft/operators/interfaces/torchkbnufft.py
paquiteau: no, non, actually I mentionend it earlier https://github.com/mind-inria/mri-nufft/pull/111#discussion_r1609956922 

You want to have the plan always in memory (just like for cufinufft). I don't know if there is some guru interface that might be more suitable

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,113 @@
...
+        Parameters
+        ----------
+        data: Tensor
+
+        Returns
+        -------
+        Tensor
+        """
+        kb_ob = self._tkb_op.forward(image=data, omega=self.samples, smaps=self.smaps)
+        return kb_ob
```
chaithyagr, line 89: I think we need safe squeeze? @paquiteau ill let you comment

chaithyagr:
@Lenoush please fix linting, you can use `ruff` and `black` for it.

src/mrinufft/operators/interfaces/torchkbnufft.py
paquiteau: the squeeze should be trigger by a `squeeze_dims` argument. I  would advise to follow what is done for the gpunufft backend in term of functionality etc.

chaithyagr:
Great work! All we need is now the tests to run for torchkbnufft. Can you please add torchkbnufft here: https://github.com/mind-inria/mri-nufft/blob/4e65f08bf3c2302ef7ffc447639a7b2ba8ffb66b/.github/workflows/test-ci.yml#L96

chaithyagr:
Please also add to pyproject.toml a new optional dependency

.vscode/settings.json
chaithyagr, line 1: This should not be commited.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+
+        B, C, XYZ = self.n_batchs, self.n_coils, self.shape
+        if not B:
+            B = 1
+        if not C:
+            C = 1
+        data = data.reshape((B, 1 if self.uses_sense else C, *XYZ))
+
+        if ktraj.shape[0] != data.shape[0]:
+            ktraj = ktraj.permute(1, 0)
```
paquiteau, line 130: why ?

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        Parameters
+        ----------
+        data: Tensor
+
+        Returns
+        -------
+        Tensor
+        """
+        ktraj = self.samples
+        smaps = self.smaps
```
paquiteau, line 151: same as op comment.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        B, C, K, XYZ = self.n_batchs, self.n_coils, self.n_samples, self.shape
+        if not B:
+            B = 1
+        if not C:
+            C = 1
```
paquiteau, line 156: same

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        data = data.reshape((B, C, K))
+
+        if ktraj.shape[0] != data.shape[0]:
+            ktraj = ktraj.permute(1, 0)
```
paquiteau, line 160: same

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        if self.density:
+            data = data * self.density
+
+        if smaps is not None:
+            smaps = smaps.to(data.dtype)
+
```
paquiteau, line 166: same

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        img = self._tkb_adj_op.forward(data=data, omega=ktraj, smaps=smaps)
+        img = img.reshape((B, 1 if self.uses_sense else C, *XYZ))
+        img /= self.norm_factor
+
+        return self._safe_squeeze(img)
+
+    @with_torch
+    def data_consistency(self, data, obs_data):
```
paquiteau, line 174: with_torch does not support the two arguments, best is to resort to manual checking here.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+            torch.is_tensor(density)
+            or isinstance(density, np.ndarray)
+            or isinstance(density, cp.ndarray)
+        ):
+            self.density = density
+        else:
+            raise ValueError(
+                "argument `density` of type" f"{type(density)} is invalid."
+            )
+
```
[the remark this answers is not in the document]
Lenoush: ok, 
so i create pipe() to compute density compensation, do I have to test it ?

src/mrinufft/operators/interfaces/torchkbnufft.py
Lenoush: the condition is useless yes :+1:

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,202 @@
...
+        if not B:
+            B = 1
+        if not C:
+            C = 1
+        data = data.reshape((B, 1 if self.uses_sense else C, *XYZ))
+
+        if ktraj.shape[0] != data.shape[0]:
+            ktraj = ktraj.permute(1, 0)
+        if smaps is not None:
+            smaps = smaps.to(data.dtype)
```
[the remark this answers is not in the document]
Lenoush: have to be dtype.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,248 @@
...
+        -------
+        Tensor: Non-uniform Fourier transform of the input image.
+        """
+        B, C, XYZ = self.n_batchs, self.n_coils, self.shape
+        data = data.reshape((B, 1 if self.uses_sense else C, *XYZ))
+
+        if self.smaps is not None:
+            self.smaps = self.smaps.to(data.dtype, copy=False)
+
+        samples_op = self.samples
```
Lenoush, line 104: I made an alias because the shape of samples must be [ndim, nlenght] only in Torchkbnufft functions but not outside them, for example in the test the shape must be [nlenght, ndim]. So transposing really can't be done in the init.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,248 @@
...
+        if self.check_samples_shape(samples_op):
```
Lenoush, line 105: My first idea was to do transposing before and after "self._tkb_op.forward" but not necessary if alias is set and it's prettier.

src/mrinufft/operators/interfaces/torchkbnufft.py
Lenoush: I don't know if i can do it like this , I left @with_torch to control "data" type and I control "obs_data" in the present function .

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,248 @@
...
+            samples_op = torch.transpose(samples_op, 1, 0)
+        kdata = self._tkb_op.forward(image=data, omega=samples_op, smaps=self.smaps)
```
chaithyagr, line 107:
```suggestion
        kdata = self._tkb_op.forward(image=data, omega=self.samples.t(), smaps=self.smaps)
```

We can keep things simple here. We know that samples will always be N x D. Also we know that omega is D X N. We dont need additional shape testing functions etc. This way, on the surface the API is the same.

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,248 @@
...
+
+        if self.smaps is not None:
+            self.smaps = self.smaps.to(data.dtype)
+
+        samples_adj_op = self.samples
+        if self.check_samples_shape(samples_adj_op):
+            samples_adj_op = torch.transpose(samples_adj_op, 1, 0)
+        img = self._tkb_adj_op.forward(
+            data=data, omega=samples_adj_op, smaps=self.smaps
+        )
```
chaithyagr, line 138: Same as above

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,248 @@
...
+        Returns
+        -------
+        bool : True if the samples shape is (ndim, klength).
+                False if the samples shape is (klength, ndim).
+
+        """
+        if samples.shape[0] == len(self.shape):
+            return False
+        else:
+            return True
```
chaithyagr, line 248:
```suggestion
```

chaithyagr (changes requested):
Some minor ways to keep things simple

src/mrinufft/operators/interfaces/torchkbnufft.py
Lenoush: ok simpler et prettier

paquiteau (approved):

chaithyagr:
Gosh torchkbnufft is reallllly slow.. We might have to rethink and maybe reduce some tests @paquiteau ?

chaithyagr:
I just double checked, it turns out that torchkbnufft is running on CPU and hence slow. Let me fix that

chaithyagr:
I seem to have opened a pandoras box. @paquiteau @Lenoush we could discuss this once. I added a parameter `use_gpu` for now.

chaithyagr (approved):
LGTM. @paquiteau to do final review and we can merge.

src/mrinufft/operators/base.py
```diff
@@ -279,7 +308,10 @@ def compute_smaps(self, method=None):
             Note that this callable function should also hold the k-space data
             (use funtools.partial)
         """
-        if isinstance(method, np.ndarray):
+        if isinstance(method, np.ndarray) or (
+            (self.backend == "torchkbnufft-cpu" or self.backend == "torchkbnufft-gpu")
+            and isinstance(method, torch.Tensor)
+        ):
```
paquiteau, line 314: would rather do the check in the operator's init (and avoid the backend specific checks here)
```python 
# in __init__ of TorchKBnufft: 
if isinstance(smaps, torch.Tensor):
   self.smaps = smaps 
else: 
   self.compute_smaps(...) 
   self.smaps = torch.from_numpy(self.smaps) 
```

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,281 @@
...
+        self.n_batchs = n_batchs
+        self.squeeze_dims = squeeze_dims
+        # self.eps = eps
+        self.compute_density(density)
+        self.compute_smaps(smaps)
+        if isinstance(smaps, np.ndarray) or (
+            CUPY_AVAILABLE and isinstance(smaps, cp.ndarray)
+        ):
+            self.smaps = torch.tensor(smaps).to("cuda" if use_gpu else "cpu")
+
```
paquiteau, line 104: To adapt (see comment on base.py)

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,281 @@
...
+        -------
+        Tensor
+        """
+        B, C, K, XYZ = self.n_batchs, self.n_coils, self.n_samples, self.shape
+        data = data.reshape((B, C, K))
+        if self.use_gpu:
+            data = data.to("cuda", copy=False)
+
+        if self.smaps is not None:
+            self.smaps = self.smaps.to(data.dtype)
```
paquiteau, line 154:
```suggestion
            self.smaps = self.smaps.to(data.dtype, copy=False)
```

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,281 @@
...
+        -------
+        Tensor: Non-uniform Fourier transform of the input image.
+        """
+        B, C, XYZ = self.n_batchs, self.n_coils, self.shape
+        data = data.reshape((B, 1 if self.uses_sense else C, *XYZ))
+        if self.use_gpu:
+            data = data.to("cuda", copy=False)
+
+        if self.smaps is not None:
+            self.smaps = self.smaps.to(data.dtype)
```
paquiteau, line 128:
```suggestion
            self.smaps = self.smaps.to(data.dtype), copy=False
```

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -0,0 +1,281 @@
...
+        data: Tensor
+
+        Returns
+        -------
+        Tensor: Non-uniform Fourier transform of the input image.
+        """
+        B, C, XYZ = self.n_batchs, self.n_coils, self.shape
+        data = data.reshape((B, 1 if self.uses_sense else C, *XYZ))
+        if self.use_gpu:
+            data = data.to("cuda", copy=False)
```
paquiteau, line 125: to avoid the various if checks, I would recommend instead:
- a `device` attribute in the init (in place of use_gpu). That could only be "cpu" or "cuda" (or a torch.device object). this also leaves the door open for multi GPU support.

paquiteau (changes requested):
This keeps on improving! A few details to make it perfect.

chaithyagr:
@paquiteau perhaps we merge this but by disabling the torchkbnufft-cpu test? Its too slow and it will hog on the CPU time we have. I think Torch can guarantee us that the codes should work well, and anyway the usecase if mostly for GPU as for CPU, finufft is the best alternative.

paquiteau (approved):

paquiteau:
:partying_face:  Congrats @Lenoush !

PR #112, opened 2024-05-22
fix: avoid copy when casting with astype.

### LINKED ISSUE #108: Calls with cupy arrays is creating unwanted extra memories.
Interface observed: `gpuNUFFT`, could potentially impact others also as it seems cupy issue

I recently saw nearly double to triple the usage of GPU memory than what was needed. On closely checking, I think the problem is when we try to convert a cupy array to `complex64`

https://github.com/mind-inria/mri-nufft/blob/8e15f052773295eee3224dd2455805d43ab11525/src/mrinufft/operators/interfaces/gpunufft.py#L311

This causes creation of new unwanted memory. We need a better converter, which should either delete the earlier memory or not even create one in the ideal scenario that we give the right datasizes. I feel we need a to complex64 as a specific separate utils function to handle this.

paquiteau:
Fixes #108

### DIFF
--- a/src/mrinufft/_utils.py
+++ b/src/mrinufft/_utils.py
@@ -52,9 +52,9 @@ def get_array_module(array):
 def auto_cast(array, dtype: DTypeLike):
     module = get_array_module(array)
     if module.__name__ == "torch":
-        return array.to(NP2TORCH[np.dtype(dtype)])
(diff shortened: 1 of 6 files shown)


chaithyagr (approved):
LGTM, merge when green

PR #114, opened 2024-05-23
Change density to density_compensators

Lenoush:
change : tfmri.estimate_density( samples, shape, method="pipe", max_iter=15)
to : tf.math.reciprocal_no_nan(tfmri.estimate_density( samples, shape, method="pipe", max_iter=15 ))

- need to add test

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -6,7 +6,12 @@ on:
   pull_request:
     branches: [ "master" ]
 
+
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 env:
   PYTHON_VERSION: "3.10"
--- a/.github/workflows/test-ci.yml
+++ b/.github/workflows/test-ci.yml
@@ -53,7 +53,7 @@ jobs:
         shell: bash
         run: |
           sudo apt install -y libnfft3-dev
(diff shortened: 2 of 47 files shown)


paquiteau:
Don't worry too much about the density tests, density compensations methods will probably need a new tests setup in the future.

chaithyagr:
This can be merged, @paquiteau ?

paquiteau:
Maybe we should add the same normalization as you did for gpuNUFFT ?

chaithyagr:
I was just going through the codes, we seem to have redundant similar pipe code at: 
https://github.com/mind-inria/mri-nufft/blob/f49c8c001ffff41d24d5267bea9c7f162b669dfb/src/mrinufft/operators/interfaces/tfnufft.py#L139-L157

@Lenoush maybe can you do the same update there also and then make the call to the `pipe` class method? See what is done in `gpuNUFFT` to get clarity.



Also can you please also implement the normalization like it is done at : 
https://github.com/chaithyagr/mri-nufft/blob/45bc400b2cfec8abee4c550780f7a6e5f1fa433c/src/mrinufft/operators/interfaces/gpunufft.py#L512-L517

paquiteau (changes requested):
Almost there ! This keeps getting better.

chaithyagr (approved):
LGTM, We can merge this right @paquiteau ?

paquiteau (approved):

Issue #104, opened 2024-05-16
A simple example to showcase autodiff


chaithyagr:
We could have possibly the following examples:
- [ ] A simple U-Net (pulled from fastMRI dataset possibly), trained? @Lenoush can you handle this in your free time?
- [ ] A network built with DeepInv library for some already preset reconstruction (PnP possibly?) @matthieutrs can you handle this this in your free time? I think you can also have something basic passed to @Lenoush and she can work on it forward.
- [x] A simple example where we could learn the trajectory possibly? @alineyyy can you handle this in your free time? #167

paquiteau:
Beware, there is also autodiff for the image/k-space data (the one that is only currently supported) This could also be show cased.

chaithyagr:
I really seemed to have clicked open issue too soon. I had written so much more xD. Lemme update

chaithyagr:
@AsmaTANABEN , can you have a look at the notebook shared in mattermost and see if we can get this on mri-nufft?

chaithyagr:
Reopening as it got closed by mistake :P

chaithyagr:
I added that we could have additional support for learning a sampling pattern in 2D with straight line readouts. 

Perhaps we can do it while supporting stackNUFFT on torch? @paquiteau is this possible? Can we pass the backprop gradients through the stacked NUFFT from the internal plane NUFFT?

chaithyagr:
On second thoughts, we would need 176 2D NUFFT then `num_samples` number of FFTs, I think a simple straightforward 3D NUFFT can be easier. @paquiteau to comment. I think in your usecase for fMRI it is more useful.

paquiteau:
after rethinking about it, stacked nufft does not have autodiff wrt to trajectory. It does support the autodiff wrt to data however.

chaithyagr:
I think the only issue left here is to do is based on deep inverse. I feel we should open a different issue for writing example to do PnP reconstructions.

PR #116, opened 2024-05-24
Autodiff wrt to trajectory

### LINKED ISSUE #103: Add support of autodiff wrt trajectory
The main idea is to implement the differentation of NUFFT forward and adjoint operator with respect to k-space trajectory locations.

Some earlier implementations:

1) TorchKBNUFFT: https://github.com/mmuckley/torchkbnufft/blob/main/torchkbnufft/_autograd/interp.py
2) TFKBNUFFT: https://github.com/zaccharieramzi/tfkbnufft/blob/master/tfkbnufft/kbnufft.py
3) tensorflow-nufft: https://github.com/mrphys/tensorflow-nufft/blob/7d2f502c6f5f4ba305e7a389ef8794b65a5b4f29/tensorflow_nufft/python/ops/nufft_ops.py#L126-L232

1) and 2) are very slow and inefficient. 3) while is much faster, it is based on Google_CUDA and reimplementation of cufinufft.

When implemented in mri-nufft, we expect a much better interface with all the backends, giving a much faster, efficient and flexible way to compute. 

Underlying math: https://web.eecs.umich.edu/~fessler/papers/files/jour/23/wang-23-eao.pdf

alineyyy:
This resolves #103


paquiteau:
@matthieutrs if you want to have a look ;)

chaithyagr (changes requested):
Thank you for great work! Not yet done with review, but a couple of issues to fix

tests/test_autodiff.py
```diff
@@ -46,6 +46,59 @@ def operator(kspace_loc, shape, backend, autograd):
     return nufft
 
 
+def proper_trajectory_torch(trajectory, normalize="pi"):
```
chaithyagr, line 49: @paquiteau this seems redundant code. Maybe we should discuss how to handle torch tensors as trajectories.

tests/test_autodiff.py
```diff
@@ -46,6 +46,59 @@ def operator(kspace_loc, shape, backend, autograd):
...
+        )
+        new_traj /= 2 * torch.pi
+
+    if normalize == "unit" and torch.max(new_traj) >= 0.5:
+        new_traj = (new_traj + 0.5) % 1 - 0.5
+
+    return new_traj
+
+
+def get_fourier_matrix_torch(ktraj, shape, dtype=torch.complex64, normalize=False):
```
chaithyagr, line 74: This too.. I think we need a better way to handle this in tests and this could potentially open new issues. We need the trajectory to be a tensor, can mri-nufft handle it? Even if it is on GPU?

chaithyagr (commented):
A couple of things to be discussed generally @paquiteau

tests/test_autodiff.py
```diff
@@ -46,6 +46,59 @@ def operator(kspace_loc, shape, backend, autograd):
     return nufft
 
 
+def proper_trajectory_torch(trajectory, normalize="pi"):
+    """Normalize the trajectory to be used by NUFFT operators on device."""
```
chaithyagr, line 50: Can be moved to `proper_trajectory`. Add additional args.

chaithyagr:
@paquiteau I handled the minor details.. locally my tests are passing. Ill add you to officially review it in free time. @alineyyy will handle your review comments. I shall try to see if I can support this in gpuNUFFT in meantime. Great job everyone!

chaithyagr:
Ahhh I forgot, @alineyyy you need to do the same updates to finufft interface...

chaithyagr:
Also, I think we need to add a test where the operator also has sensitivity maps. I am unsure if we shall pass tests with it, but I think we should technically. But its worth testing it.

chaithyagr:
I have successfully added suppoprt for autodiff wrt to gpuNUFFT! Note that you need to use gpuNUFFT 0.8.0 to be able to support!

src/mrinufft/operators/base.py
```diff
@@ -103,11 +104,11 @@ class or instance of class if args or kwargs are given.
...
-    if autograd:
-        if isinstance(operator, FourierOperatorBase):
-            operator = operator.make_autograd(variable=autograd)
-        else:  # partial
-            operator = partial(operator.with_autograd, variable=autograd)
+    # if autograd:
+    if isinstance(operator, FourierOperatorBase):
+        operator = operator.make_autograd(wrt_data, wrt_traj)
+    else:  # partial
+        operator = partial(operator.with_autograd, wrt_data, wrt_traj)
```
chaithyagr, line 111:
```suggestion
    elif (wrt_data or wrt_traj):  # partial
        operator = partial(operator.with_autograd, wrt_data, wrt_traj)
```


With this, we enable the `make_autograd` pipeline only if needed. This way you wont need `torch` always

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,42 @@
...
+        r = [torch.linspace(-s / 2, s / 2 - 1, s, device=device) for s in shape]
+        grid_r = torch.meshgrid(r, indexing="ij")
+        grid_r = torch.reshape(torch.stack(grid_r), (ndim, n)).to(device)
+        traj_grid = torch.matmul(ktraj, grid_r)
+        matrix = torch.exp(-2j * np.pi * traj_grid).to(dtype).to(device).clone()
+    else:
+        r = [np.linspace(-s / 2, s / 2 - 1, s) for s in shape]
+        grid_r = np.reshape(np.meshgrid(*r, indexing="ij"), (ndim, np.prod(shape)))
+        traj_grid = ktraj @ grid_r
+        matrix = np.exp(-2j * np.pi * traj_grid, dtype=dtype)
```
chaithyagr, line 31: This can be refactored more:

To make it cleaner, maybe have `xp = get_array_module`, then use xp.

```python
    r = [module.linspace(-s / 2, s / 2 - 1, s, dtype=dtype) for s in shape]
    grid_r = module.meshgrid(r, indexing="ij")
    grid_r = module.reshape(torch.stack(grid_r), (ndim, n))
    if module.__name__ == 'torch':
        grid_r = grid_r.to(device)
    traj_grid = module.matmul(ktraj, grid_r)
    matrix = module.exp(-2j * np.pi * traj_grid)
```


Dont commit this code as is, its just a suggestion. This way we can prevent a lot of repeations

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,42 @@
...
     if normalize:
-        matrix /= np.sqrt(np.prod(shape)) * np.power(np.sqrt(2), len(shape))
+        matrix /= (
+            (
+                torch.sqrt(torch.tensor(np.prod(shape), device=device))
+                * torch.pow(torch.sqrt(torch.tensor(2, device=device)), ndim)
+            )
+            if module.__name__ == "torch"
+            else (np.sqrt(np.prod(shape)) * np.power(np.sqrt(2), len(shape)))
+        )
```
chaithyagr, line 41: This could potenitally be a problem if someone uses `normalize=True` with numpy! Try to follow using module here to prevent that

tests/operators/test_bindings.py
```diff
@@ -18,7 +18,7 @@
 )
 def test_get_operator(backend, name):
     """Test the get_operator function."""
-    assert mrinufft.get_operator(backend).__name__ == name
+    assert mrinufft.get_operator(backend).func.__self__.__name__ == name
```
chaithyagr, line 21: You will not need this after my above change in `get_operator`

tests/test_autodiff.py
```diff
@@ -23,54 +22,66 @@
...
         shape=shape,
         smaps=None,
         squeeze_dims=False,  # Squeezing breaks dimensions !
     )
-
     return nufft
 
 
-@fixture(scope="module")
+# @fixture(scope="module")
```
chaithyagr, line 46:
```suggestion
```

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -89,12 +90,31 @@ def _make_plan(self, typ, **kwargs):
...
+                cp.array(self.samples[:, 0], copy=False),
+                cp.array(self.samples[:, 1], copy=False),
+                cp.array(self.samples[:, 2], copy=False) if self.ndim == 3 else None,
+            )
+        else:
+            self.plans[typ].setpts(
+                cp.array(self.samples[:, 0], copy=False),
+                cp.array(self.samples[:, 1], copy=False),
+                cp.array(self.samples[:, 2], copy=False) if self.ndim == 3 else None,
+            )
```
chaithyagr, line 117:
```suggestion
        plan = self.grad_plan if typ == "grad" else self.plans[typ] 
        plan.setpts(
            cp.array(self.samples[:, 0], copy=False),
            cp.array(self.samples[:, 1], copy=False),
            cp.array(self.samples[:, 2], copy=False) if self.ndim == 3 else None,
        )
```

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -89,12 +90,31 @@ def _make_plan(self, typ, **kwargs):
...
 
     def _destroy_plan(self, typ):
         if self.plans[typ] is not None:
```
chaithyagr, line 120: Perhaps, you might need to also need to write a way to destroy grad_plan

src/mrinufft/operators/interfaces/finufft.py
```diff
@@ -47,11 +48,29 @@ def _make_plan(self, typ, **kwargs):
...
+        if typ == "grad":
+            fpts_axes = [None, None, None]
+            for i in range(self.ndim):
+                fpts_axes[i] = np.array(self.samples[:, i], dtype=self.samples.dtype)
+            self.grad_plan.setpts(*fpts_axes)
+        else:
+            fpts_axes = [None, None, None]
+            for i in range(self.ndim):
+                fpts_axes[i] = np.array(self.samples[:, i], dtype=self.samples.dtype)
+            self.plans[typ].setpts(*fpts_axes)
```
chaithyagr, line 73: Please make same change as I proposed above to keep code cleaner

chaithyagr (changes requested):
Great work! Thank you @alineyyy for your conteributions. A bunch of comments to be handeled. Also, I think we need to update the test_ci.yml file. I think `test_autodiff` is never run. Ill have a look around and get back

chaithyagr:
I moved `test_autodiff` to `operators` so that it runs also. @paquiteau JFYI...

src/mrinufft/operators/interfaces/nudft_numpy.py
alineyyy: Here, I think my code isn't concise yet, because I found that we have to put the operations to device when torch and it can't be executed when numpy...Do you have some ideas to solve this problem??

src/mrinufft/operators/interfaces/cufinufft.py
alineyyy: > Perhaps, you might need to also need to write a way to destroy grad_plan

I have added this function in cufinufft, but I am not sure if I need to call this function in any other files...

src/mrinufft/operators/interfaces/nudft_numpy.py
alineyyy: > This can be refactored more:
> 
> To make it cleaner, maybe have `xp = get_array_module`, then use xp.
> 
> ```python
>     r = [module.linspace(-s / 2, s / 2 - 1, s, dtype=dtype) for s in shape]
>     grid_r = module.meshgrid(r, indexing="ij")
>     grid_r = module.reshape(torch.stack(grid_r), (ndim, n))
>     if module.__name__ == 'torch':
>         grid_r = grid_r.to(device)
>     traj_grid = module.matmul(ktraj, grid_r)
>     matrix = module.exp(-2j * np.pi * traj_grid)
> ```
> 
> Dont commit this code as is, its just a suggestion. This way we can prevent a lot of repeations

Here, I think my code isn't concise yet, because I found that we have to put the operations to device when torch and it can't be executed when numpy...Do you have some ideas to solve this problem??

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,41 @@
...
+
     if normalize:
-        matrix /= np.sqrt(np.prod(shape)) * np.power(np.sqrt(2), len(shape))
+        norm_factor = (
+            module.sqrt(module.prod(module.tensor(shape, device=device)))
+            * module.pow(module.sqrt(module.tensor(2, device=device)), ndim)
+            if module.__name__ == "torch"
+            else (module.sqrt(module.prod(shape)) * module.power(module.sqrt(2), ndim))
+        )
+        matrix /= norm_factor
```
chaithyagr, line 40: @paquiteau , ideas to have the code cleaner?

chaithyagr (approved):
LGTM, Adding @paquiteau for some final checks and then we can merge when green.

src/mrinufft/operators/base.py
```diff
@@ -103,11 +104,11 @@ class or instance of class if args or kwargs are given.
...
 
-    if autograd:
-        if isinstance(operator, FourierOperatorBase):
-            operator = operator.make_autograd(variable=autograd)
-        else:  # partial
-            operator = partial(operator.with_autograd, variable=autograd)
+    # if autograd:
+    if isinstance(operator, FourierOperatorBase):
+        operator = operator.make_autograd(wrt_data, wrt_traj)
+    elif wrt_data or wrt_traj:  # partial
```
paquiteau, line 110:
```suggestion
    elif wrt_data or wrt_traj:  # instance will be created later
```

src/mrinufft/operators/base.py
```diff
@@ -290,12 +291,12 @@ def compute_smaps(self, method=None):
...
 
-    def make_autograd(self, variable="data"):
+    def make_autograd(self, wrt_data=True, wrt_traj=False):
         """Make a new Operator with autodiff support.
 
         Parameters
         ----------
-        variable: str, default data
+        variable: , default data
             variable on which the gradient is computed with respect to.
```
paquiteau, line 301: Update the doc

src/mrinufft/operators/base.py
```diff
@@ -310,10 +311,8 @@ def make_autograd(self, variable="data"):
         """
         if not AUTOGRAD_AVAILABLE:
             raise ValueError("Autograd not available, ensure torch is installed.")
```
paquiteau, line 313: This should check that the backend supports autodiff. 
Ideally, this should be a class attribute defined for each backend (by default False in the `FourierOperatorBase`. I would suggest something along these lines:

```python 
class FourierOperatorBase: 
     ... 
     autograd_available = False 

# in each backend that support autograd, e.g. 
class MRICufinufft(FourierOperatorBase):
      backend = "cufinufft"
      autograd_available = True 
```

Then, in this function we will test that both AUTOGRAD_AVAILABLE and self.autograd_available are True. Otherwise, raise an error (with a specific error message in each case)

src/mrinufft/operators/base.py
```diff
@@ -485,9 +484,9 @@ def __repr__(self):
         )
 
     @classmethod
-    def with_autograd(cls, variable, *args, **kwargs):
+    def with_autograd(cls, wrt_data, wrt_traj, *args, **kwargs):
```
paquiteau, line 487: Add a docstring, and default values for the arguments. 
```suggestion
    def with_autograd(cls, *args, wrt_data=True, wrt_traj=False, **kwargs):
```

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,41 @@
...
 import numpy as np
 import scipy as sp
-
 from ..base import FourierOperatorCPU
-from mrinufft._utils import proper_trajectory
+from mrinufft._utils import proper_trajectory, get_array_module
 
 
 def get_fourier_matrix(ktraj, shape, dtype=np.complex64, normalize=False):
     """Get the NDFT Fourier Matrix."""
```
paquiteau, line 12: some documentation as well if possible ?

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,41 @@
...
+    module = get_array_module(ktraj)
```
paquiteau, line 13: module -> xp for consistency with the rest of the codebase

src/mrinufft/operators/interfaces/nudft_numpy.py
paquiteau: The ternary is too long, use  a regular if-else block.
Also, the norm factor is just a scalar, everything can be done in numpy regardless of the application afterwards.

tests/case_trajectories.py
```diff
@@ -51,6 +51,12 @@ def case_nyquist_radial3D(self, Nc=32 * 4, Ns=16, Nr=32 * 4, N=32):
         trajectory = rotate(trajectory, nb_rotations=Nr)
         return trajectory, (N, N, N)
 
+    def case_nyquist_radial3D_lowmem(self, Nc=2, Ns=16, Nr=2, N=10):
+        """Create a 3D radial trajectory."""
```
paquiteau, line 55:
```suggestion
        """Create a 3D radial trajectory with low memory."""
```

tests/operators/test_autodiff.py
```diff
@@ -23,54 +22,65 @@
...
         samples=kspace_loc,
         shape=shape,
         smaps=None,
         squeeze_dims=False,  # Squeezing breaks dimensions !
     )
-
     return nufft
 
 
-@fixture(scope="module")
```
paquiteau, line 49: Is there a reason to move away from the fixture setup?

paquiteau (changes requested):
This is a major a milestone for MRI-NUFFT, and we are never been so close of merging. Thanks @alineyyy for making this possible !

Some comments/questions to address, and a few suggestion for making things even better.

tests/operators/test_autodiff.py
chaithyagr: Wait this is new to me.. I have no idea... @alineyyy if you are unaware, please add it back..

alineyyy: Ohhhhh sorry!! I didn't notice that and I will add it back

alineyyy: For this fixture setup, I find that I can't add it back : ( 
I think operator.samples(trajectory) need to be put on device before we get the NDFT matrix. If we want to transform the data to device in function ndft_matrix, we need to add a parameter "interface" which will lead to a failed execution of the tests. 
Do you have any ideas to change here in order to add the setup back? @paquiteau @chaithyagr

chaithyagr: Thats good catch. I am okay with current state (as long as we run the test), Ill let @paquiteau if he has a better fix

tests/operators/test_autodiff.py
```diff
@@ -81,22 +91,38 @@ def test_adjoint_and_grad(operator, ndft_matrix, interface):
...
+def test_forward_and_grad(operator, interface):
     """Test the adjoint and gradient of the operator."""
     if operator.backend == "finufft" and "gpu" in interface:
         pytest.skip("GPU not supported for finufft backend")
 
-    ndft_matrix_torch = to_interface(ndft_matrix, interface=interface)
+    if torch.is_tensor(operator.samples):
+        operator.samples = operator.samples.cpu().detach().numpy()
+
+    operator.samples = to_interface(operator.samples, interface=interface)
```
chaithyagr, line 102: Following on to my previous comment. Something to think about, this makes using with autodiff slightly complicated as users cannot just do:

trajectory = torch.Tensor(...)
trajectory.requires_grad = True

get_operator('cufinufft', grad_wrt_traj=True)(trajectory)


and then get gradients. Right? I think we ought to support that. Need not be in this PR. Maybe when we are getting example in place in #104 , we better have a simpler version of codes. @alineyyy @paquiteau What do you think?

src/mrinufft/operators/interfaces/finufft.py
```diff
@@ -47,11 +48,24 @@ def _make_plan(self, typ, **kwargs):
             **kwargs,
         )
 
+    def _make_plan_grad(self, **kwargs):
+        self.grad_plan = Plan(
+            2,
+            self.shape,
+            self.n_trans,
+            self.eps,
+            dtype="complex64" if self.samples.dtype == "float32" else "complex128",
```
paquiteau, line 57: you can use `DTYPE_R2C` dict for that

src/mrinufft/operators/interfaces/nudft_numpy.py
```diff
@@ -4,23 +4,55 @@
...
+        r = [x.to(device) for x in r]
+
+    grid_r = xp.meshgrid(*r, indexing="ij")
+    grid_r = xp.reshape(xp.stack(grid_r), (ndim, n))
+    traj_grid = xp.matmul(ktraj, grid_r)
+    matrix = (
+        xp.exp(-2j * xp.pi * traj_grid).to(dtype).to(device).clone()
+        if xp.__name__ == "torch"
+        else (xp.exp(-2j * xp.pi * traj_grid, dtype=dtype))
+    )
```
paquiteau, line 48:
```suggestion
    matrix = xp.exp(-2j * xp.pi * traj_grid)
    if xp.__name__ == "torch":
        matrix.to(dtype=dtype, device=device, copy=True)
    )
```

paquiteau (changes requested):
almost there, there are a few comments to address still.

src/mrinufft/operators/interfaces/nudft_numpy.py
alineyyy: I think I have to use matrix = matrix.to(...) here, otherwise it will occur an error in the autograd function of pytorch.

tests/operators/test_autodiff.py
paquiteau: Yes. We will track this in #127

paquiteau (approved):

chaithyagr:
If the failing gpunufft test is too strict causing problems, just ensure the gradients are close and maybe just make the test less strict

PR #119, opened 2024-05-27
Some fixes in Smaps blurr function and slice and contrast based stuff

chaithyagr:
We ideally will move the codes of rawdat file reader to pysap-mri eventually. BUt adding here for now.


src/mrinufft/operators/base.py
```diff
@@ -542,6 +542,7 @@ def __init__(
 
         # Density Compensation Setup
         self.compute_density(density)
+        self.compute_smaps(smaps)
         # Multi Coil Setup
         self.compute_smaps(smaps)
```
paquiteau, line 547: this is duplicated code it seems.

paquiteau (changes requested):
Some cleanup to do. 

I would also advise to document the blurring of the smaps.

src/mrinufft/operators/base.py
chaithyagr: Thats weird, good catch

Issue #132, opened 2024-06-21
Update the trajectory and DCp when self.samples is updated


chaithyagr:
This issue will track a bunch of features needed for learning k-space sampling trajectories.

For now, what we need is:
- [x] `self.samples` property setter to have update the corresponding plans to update the trajectory in respective `raw_op` for each backend. 
     - [x] This is doable in `finufft` and `cufinufft` easily as there is a set_pts function. @alineyyy will handle a PR for it. @alineyyy for other backends, please add an error for now. This can be done by having error raised in the main setter in the base class. 
     - [X] For `gpuNUFFT`, we need some update on `gpuNUFFT` side. @chaithyagr will handle this. #133 
     - [x] @paquiteau can you detail more about other backends and if want to support for them? (Most dont have autodiff wrt to trajectory anyway).
- [x] Update the density compensators accordingly. This is a bit complicated as it means we need to save how the density was estimated. @paquiteau lets discuss about this and @Lenoush can you please work on it?
- [X] A bunch of tests overall to test this. This is fairly straightforward. But its good sanity tests #133

chaithyagr:
All of this is handeled now!

PR #133, opened 2024-06-21
Add support for updating points with gpuNUFFT

chaithyagr:
This handles a part of #132 . Also added tests, which can be extended by @alineyyy to test her use case.
Adding @paquiteau for review. Notice that we need more recent version of gpuNUFFT for it to pass. Tests passing locally for now, but PyPi release is also done, so we can merge ONLY when all tests are green.


tests/case_trajectories.py
```diff
@@ -51,7 +51,7 @@ def case_nyquist_radial3D(self, Nc=32 * 4, Ns=16, Nr=32 * 4, N=32):
         trajectory = rotate(trajectory, nb_rotations=Nr)
         return trajectory, (N, N, N)
 
-    def case_nyquist_radial3D_lowmem(self, Nc=2, Ns=16, Nr=2, N=10):
+    def case_nyquist_lowmem_radial3D(self, Nc=2, Ns=16, Nr=2, N=10):
```
chaithyagr, line 54: This was a sidequest, the batch tests were triggering to run on this also, not like it matters, but just saving some time by changing the name

tests/operators/test_update.py
```diff
@@ -0,0 +1,139 @@
...
+        (3, 4, 2, True),
+        (3, 4, 2, False),
+    ],
+)
+@parametrize_with_cases(
+    "kspace_locs, shape",
+    cases=CasesTrajectories,
+    glob="*random*",
+)
+@parametrize(backend=["gpunufft"])
```
chaithyagr, line 40: @alineyyy you can add `finufft` and `cufinufft` here after you add similar support in it. (Remember to also update points for the grad_plan)

src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -481,6 +500,23 @@ def uses_sense(self):
...
+        ----------
+        samples: np.ndarray
+            The samples for the Fourier Operator.
+        """
+        if self.density_method is not None:
+            self.compute_density(self.density_method)
+        self.raw_op.set_pts(
+            samples,
+            density=self.density,
+        )
```
paquiteau, line 517: At initialization this may triggers errors (raw_op might not be defined), and if it is defined, this does an extra computation.  
Maybe a `.update_samples` would be more convenient.

tests/case_trajectories.py
paquiteau: wait, renaming the function make it run faster ?

paquiteau (changes requested):
Maybe overcharging the setter is a bad idea.

chaithyagr:
Adding what we discussed for documenting. We did overcharge the setter as we want to ensure that everything wrt to NUFFT like density compensation update can happen in background when a train function is called. 
Also @paquiteau , in future, other trajectory learning can have a projector within the setter function.

paquiteau (approved):

PR #134, opened 2024-06-21
Numpy2.0

paquiteau:
Numpy 2.0 got released, some nufft backend are broken, this fixes them in the CI.

### DIFF
--- a/.github/workflows/test-ci.yml
+++ b/.github/workflows/test-ci.yml
@@ -53,6 +53,7 @@ jobs:
         shell: bash
         run: |
           sudo apt install -y libnfft3-dev
+          python -m pip install numpy<2.0.0
           python -m pip install pynfft2
 
       - name: Install pynufft
--- a/src/mrinufft/operators/base.py
+++ b/src/mrinufft/operators/base.py
@@ -109,10 +109,12 @@ def get_operator(
         operator = operator(*args, **kwargs)
 
     # if autograd:
-    if isinstance(operator, FourierOperatorBase):
-        operator = operator.make_autograd(wrt_data, wrt_traj)
-    elif wrt_data or wrt_traj:  # instance will be created later
-        operator = partial(operator.with_autograd, wrt_data, wrt_traj)
+    if wrt_data or wrt_traj:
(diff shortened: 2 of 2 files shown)


src/mrinufft/operators/base.py
```diff
@@ -109,10 +109,12 @@ class or instance of class if args or kwargs are given.
...
-    if isinstance(operator, FourierOperatorBase):
-        operator = operator.make_autograd(wrt_data, wrt_traj)
-    elif wrt_data or wrt_traj:  # instance will be created later
-        operator = partial(operator.with_autograd, wrt_data, wrt_traj)
+    if wrt_data or wrt_traj:
+        if isinstance(operator, FourierOperatorBase):
+            operator = operator.make_autograd(wrt_data, wrt_traj)
+        else:
+            # instance will be created later
+            operator = partial(operator.with_autograd, wrt_data, wrt_traj)
```
chaithyagr, line 117: Is this related to this PR?

.github/workflows/test-ci.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         shell: bash
         run: |
           sudo apt install -y libnfft3-dev
+          python -m pip install numpy<2.0.0
```
chaithyagr, line 56: shouldnt we update in pyproject.toml? Like is it a test issue or lack of compatibility?

src/mrinufft/operators/base.py
paquiteau: no, but this is a fix for something else (that I though was already merge).

.github/workflows/test-ci.yml
paquiteau: It's more complicated than that, some backend requires numpy<2.0, other don't care

chaithyagr: Maybe we document the problem in issue and tag it here? That way we know what needs a fix.. Is the pipeline broken currently?

chaithyagr, line 56:
```suggestion
```

Issue #50, opened 2023-11-15
Add TURBINE trajectory


paquiteau:
Presented in: 

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9546489/

This is not equivalent to a stack of 2D-radial trajectory due to the interleaving of the blades (see figure)

![image](https://github.com/mind-inria/mri-nufft/assets/77174042/0a33e626-3e41-4e68-a1b5-d4b2413a23ca)

Daval-G:
From my current understanding, this pattern is extremely similar to 3D-REPI and therefore I plan to deliver both under the same PR. See issue #49 on 3D-REPI for common details.

Issue #49, opened 2023-11-15
Add 3D-REPI


paquiteau:
Another non-cartesian trajectory that has some nice properties

An example: 
![image](https://github.com/mind-inria/mri-nufft/assets/77174042/3f2ea49d-6a02-403c-a633-c302ce79229a)

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8616809/

Daval-G:
Both 3D-REPI and TURBINE (from issue #50) are quite simple to implement except for one question: How should we handle split readouts ? EPI-like trajectories are characterized by blips/short transition between each line composing a single long shot. For example, the blue lines above are acquired over only one TR, but transitions between each line/spiral is not shown and could have an arbitrary duration depending on the sequence. To me it leads to two possible propositions:
1) In the trajectory initialization, we request a bunch of additional parameters for each EPI-like trajectory about how many points should be used for the transition, and we deduce the slew rate to apply in transition.
2) The EPI-like trajectories only provides lines/spirals separately but ordered implicitely as "interleaves", and we also provide a utilitary function for the user to transform those interleaves into single shots, as it would typically be used for GRE-like acquisitions. It could also be used on other trajectories that are naturally interleaved.

I prefer the second solution, it makes trajectory initialization much lighter and invite to the development of different blip/transition strategies. WDYT ?
In both cases, we will provide additional gradient manipulation techniques to create transitions, but also prewind & rewind gradients.

paquiteau:
I like the 2nd proposal a lot! Those considerations are only relevant when using the trajectories on the scanner, so providing the idealized trajectory (as it would be used in NUFFT operator initialization) and the way to make it hardware-compatible by having proper blips is probably the best way to go (and generalize well as you mention). 

I look forward to seeing your implementations!

PR #137, opened 2024-06-23
Add TURBINE/REPI with EPI-based tools

Daval-G:
This PR adds the following features:
- `initialize_3D_TURBINE` function to create a TURBINE trajectory requested in issues #50, that consists in EPI shots rotated into a cylinder,
- `initialize_3D_REPI` function to create a REPI trajectory requested in issues #49, that can be either radial or spiral EPI-like shots interlaced between blades rather than stacked independently
- `epify` function used to combine shots into a single multi-echo readout train, similar to EPI shots
- `prewind` and `rewind` to create the positioning & spoiling gradients before and after the readouts, and therefore obtain the full gradient sequence rather than just the one played during the readouts.

The different functions & features are illustrated below, with the TURBINE & REPI trajectories (up) and the epify/prewind/rewind tools applied to different trajectories (down)

Screenshot from 2024-06-23 18-15-06

Screenshot from 2024-06-23 18-15-36


paquiteau:
This looks very promising ! I will let you add documentation before reviewing. 

Some thoughts/Questions: 

- Maybe a un-epify method could be added as well if it make sense ? (you have an EPI trajectory with the gradients turns, and you want to only preserve the straight lines for the reconstruction ) 
- Does the original REPI / TURBINE support GRAPPA acceleration ? do we want to add this to MRI-NUFFT ?

Daval-G:
> This looks very promising ! I will let you add documentation before reviewing.
> 
> Some thoughts/Questions:
> 
> * Maybe a un-epify method could be added as well if it make sense ? (you have an EPI trajectory with the gradients turns, and you want to only preserve the straight lines for the reconstruction )

Yes that make sense, I will add that too

> * Does the original REPI / TURBINE support GRAPPA acceleration ? do we want to add this to MRI-NUFFT ?

Didn't read about it, but it is quite certain they can. I would suggest to have it made through another function/tool rather than keep stacking arguments on REPI/TURBINE functions, and add an example where we make them GRAPPA-compatible. It requires to add an option to not epify automatically the REPI & TURBINE functions, but I was planning to propose that anyway.

chaithyagr:
@Daval-G thank you for this work. I was curious if you had something more to be done here? If not, I can handle the linting issue.

Daval-G:
> @Daval-G thank you for this work. I was curious if you had something more to be done here? If not, I can handle the linting issue.

The examples are still missing, that's what I am currently working on

Daval-G:
@paquiteau good to go ?

Daval-G:
There is still the question of the in-plane acceleration, but I plan to do that in another PR

paquiteau (approved):

PR #138, opened 2024-06-24
Update plan in time for cufinufft and finufft

alineyyy:
This solve  the first part of #132 .
The plan of cufinufft and finufft operator can be updated in time.

### DIFF
--- a/src/mrinufft/operators/interfaces/cufinufft.py
+++ b/src/mrinufft/operators/interfaces/cufinufft.py
@@ -265,11 +265,12 @@ class MRICufiNUFFT(FourierOperatorBase):
             **kwargs,
         )
         # Support for concurrent stream and computations.
-        
+
     @FourierOperatorBase.samples.setter
-    def samples(self, samples): 
+    def samples(self, samples):
+        """Update the plans when changing the samples."""
         self._samples = samples
-        for typ in [1, 2, 'grad']:
+        for typ in [1, 2, "grad"]:
             self.raw_op._set_pts(typ)
 
     @with_numpy_cupy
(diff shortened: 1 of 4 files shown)


src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -72,6 +72,8 @@ def __init__(
             self._make_plan(i, **kwargs)
             self._set_pts(i)
 
+        self._make_plan_grad(**kwargs)
```
chaithyagr, line 75: We call this only if needed

```suggestion
```

src/mrinufft/operators/interfaces/finufft.py
```diff
@@ -40,6 +40,8 @@ def __init__(
             self._make_plan(i, **kwargs)
             self._set_pts(i)
 
+        self._make_plan_grad(**kwargs)
+
```
chaithyagr, line 44:
```suggestion
```

Same as above. We dont need additional plan which will take up more memory. What we need to do is to trigger this creation only when `grad_wrt_traj=True`

src/mrinufft/operators/interfaces/finufft.py
```diff
@@ -141,12 +149,13 @@ def __init__(
...
-            samples,
-            shape,
-            n_trans=n_trans,
-            **kwargs,
-        )
+    @FourierOperatorBase.samples.setter
+    def samples(self, samples):
+        """Update the plans when changing the samples."""
+        self._samples = samples
+        for typ in [1, 2, "grad"]:
```
chaithyagr, line 160: Now I see why you initialized _make_grad_plan above. However, we should update points for grad_plans only if grad_wrt_traj is set to True.

chaithyagr (changes requested):
Thank you for your changes! Minor issue to be handeled and then we can merge when green.

chaithyagr (approved):
As they are minor changes, I handeled them. LGTM. We can merge once #139 goes in and we re-test.

PR #142, opened 2024-06-28
A very basic working example to learn samples

chaithyagr:
This had to be setup after discussing with @alineyyy , where we found some issues due to which it was harder to instance the operator inside a torch module. @paquiteau , maybe a quick review of the code to ensure we agree on style (we can meet if required). @alineyyy please update the learn samples codes to make it cleaner. Please also add a simple projection step to ensure we sample the center of k-space.


examples/example_learn_samples.py
```diff
@@ -0,0 +1,67 @@
...
+class Model(torch.nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
+        self.operator = get_operator("gpunufft", wrt_data=True, wrt_traj=True)(
+            initialize_2D_radial(10, 192).reshape(-1, 2).astype(np.float32),
+            shape=(256, 256),
+            density=True,
+            squeeze_dims=False,
+        )
+        self.trajectory = self.operator.samples_torch
```
chaithyagr, line 36: This is just done for simplicity

src/mrinufft/_utils.py
```diff
@@ -86,7 +86,7 @@ def proper_trajectory(trajectory, normalize="pi"):
         raise ValueError(
             "trajectory should be array_like, with the last dimension being coordinates"
         ) from e
-
+
```
chaithyagr, line 89:
```suggestion
```

src/mrinufft/operators/autodiff.py
```diff
@@ -135,15 +137,25 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
...
-        return _NUFFT_ADJOP.apply(kspace, self.samples, self.nufft_op)
+        return _NUFFT_ADJOP.apply(kspace, self.samples_torch, self.nufft_op)
 
     def __getattr__(self, name):
         """Get the attribute from the root operator."""
-        return getattr(self.nufft_op, name)
+        if hasattr(self.nufft_op, name):
+            if name == 'samples':
+                return self.samples_torch
+            return getattr(self.nufft_op, name)
```
paquiteau, line 159: Don't make the exception here, just add a property. Also this will break if you don't have it initialized. 

```python
@property
def samples(self):
    try:
        return self._samples_torch 
    except AttributeError:
        return self.nufft_op.samples

def __getattr__(self, name):
    return getattr(self.nufft_op, name)
```

src/mrinufft/operators/autodiff.py
```diff
@@ -135,15 +137,25 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
...
+        else:
+            return super().__getattr__(name)
```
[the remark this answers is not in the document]
paquiteau: see above

tests/operators/test_autodiff.py
```diff
@@ -75,15 +75,14 @@ def test_adjoint_and_grad(operator, interface):
...
         pytest.skip("GPU not supported for finufft backend")
 
-    if torch.is_tensor(operator.samples):
-        operator.samples = operator.samples.cpu().detach().numpy()
-
-    operator.samples = to_interface(operator.samples, interface=interface)
+    if "gpu" in interface:
+        operator.to("cuda")
+    else:
+        operator.cpu()
```
paquiteau, line 81: what does this mean ? operator is not a tensor here.

examples/example_learn_samples.py
```diff
@@ -0,0 +1,67 @@
...
+from mrinufft import check_backend, get_density, get_operator
+from mrinufft.trajectories import initialize_2D_radial
+from mrinufft.trajectories.display import display_2D_trajectory
+
+# %%
+# Create sample data
+# ------------------
+
+class Model(torch.nn.Module):
+    def __init__(self):
```
paquiteau, line 28: Why do you need the extra module layer ?

paquiteau (changes requested):
A good start, adding an example for the autodiff is clearly required ! However I would recommend to add more detail in the example to explains what is happening.

examples/example_learn_samples.py
```diff
@@ -0,0 +1,67 @@
...
+Density Compensation Routines
+=============================
+
+Examples of differents density compensation methods.
+
+Density compensation depends on the sampling trajectory,and is apply before the
+adjoint operation to act as preconditioner, and should make the lipschitz constant
+of the operator roughly equal to 1.
+
+"""
```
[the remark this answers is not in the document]
chaithyagr: AHh no no, its just a temporary file. Let me mark this as draft

src/mrinufft/operators/autodiff.py
```diff
@@ -135,15 +137,25 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
         if wrt_traj and self.nufft_op.backend in ["finufft", "cufinufft"]:
             self.nufft_op.raw_op._make_plan_grad()
         self.nufft_op._grad_wrt_data = wrt_data
-
+        if wrt_traj:
+            self.samples_torch = torch.nn.Parameter(
+                data=torch.Tensor(self.nufft_op.samples),
+                requires_grad=True,
```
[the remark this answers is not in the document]
chaithyagr: But samples will route to samples_torch. But I understand your point, lemme do it. Its just that I am wondering if torch can handle the routing and properties right for autodiff.

src/mrinufft/operators/autodiff.py
chaithyagr: Ahhh good point, but then if we are in autodiff module with grad_wrt_traj, we SHOULD have it initialized. 
But I dont know if it misses any corner cases (perhaps if we want grad_wrt_data and request `samples`

tests/operators/test_autodiff.py
chaithyagr: `operator` is a `torch.nn.module` here. This means all the underlying parameters / weights will be on GPU or CPU. It is very "torch-like" code, and this is good for us

examples/example_learn_samples.py
chaithyagr: I think and expect this is how users will use our operator. The `test` handles the case when they dont instance in a module and the example handles it being instanced in a module. Note that this was failing earlier, and hence we have this PR to begin with. Eventually we might want the model to dynamically update the shape of the trajectory (during upsampling steps for example.

chaithyagr:
Closing as #167 is now here!

PR #156, opened 2024-07-11
Smaps update and prevent saving trajectory for autodiff

### LINKED ISSUE #155: Add a method to update the Smaps dynamically.
This can be implemented through a setter function for Smaps:

This could lead to different implementations for each interface:

- [ ] gpuNUFFT: The setter function shall update the pinned_smaps (dont create new array though, just do the copy):
https://github.com/mind-inria/mri-nufft/blob/eab0d3c4eb49cb5e6cd192c118fcf0d0e29cbb0a/src/mrinufft/operators/interfaces/gpunufft.py#L45-L48

- [ ] cufinufft: We need to check if smaps_cached is set and accordingly copy it either to GPU or a pinned array. See the init function.
- [ ] torchkbnufft: Just update the incoming smaps to a torch tensor.

Finally, a quick test in `test_update.py` module, where the Smaps is also updated could help to ensure the codes are checked.

This can be helpful to learn networks with update of smaps without needing to re-create the operator.

AsmaTANABEN:
This resolves #155 smaps update support for cufinufft and gpunufft backends, via the setter function.

Additionally, we were saving the trajectory for autodiff in the past but it is not needed as we are anyway having trajectory within the plans and the nufft_op which is saved anyway.


src/mrinufft/operators/autodiff.py
```diff
@@ -20,14 +20,20 @@ class _NUFFT_OP(torch.autograd.Function):
     @staticmethod
     def forward(ctx, x, traj, nufft_op):
         """Forward image -> k-space."""
-        ctx.save_for_backward(x, traj)
+        if nufft_op._grad_wrt_data and not nufft_op._grad_wrt_traj:
+            ctx.save_for_backward(x)
+        else :
+            ctx.save_for_backward(x, traj)
```
paquiteau, line 26: and what if you have `wrt_traj=true` but `wrt_data=False` there is 3 case to check (the 4th, all false is irrelevant)

src/mrinufft/operators/autodiff.py
```diff
@@ -20,14 +20,20 @@ class _NUFFT_OP(torch.autograd.Function):
...
         return nufft_op.op(x)
 
     @staticmethod
     def backward(ctx, dy):
         """Backward image -> k-space."""
-        (x, traj) = ctx.saved_tensors
+        if ctx.nufft_op._grad_wrt_data and not ctx.nufft_op._grad_wrt_traj:
+            x = ctx.saved_tensors
+        else:
+            (x, traj) = ctx.saved_tensors
```
paquiteau, line 36: same as above

src/mrinufft/operators/autodiff.py
```diff
@@ -72,14 +78,20 @@ class _NUFFT_ADJOP(torch.autograd.Function):
     @staticmethod
     def forward(ctx, y, traj, nufft_op):
         """Forward kspace -> image."""
-        ctx.save_for_backward(y, traj)
+        if nufft_op._grad_wrt_data and not nufft_op._grad_wrt_traj:
+            ctx.save_for_backward(y)
+        else :
+            ctx.save_for_backward(y, traj)
```
paquiteau, line 84: same

src/mrinufft/operators/autodiff.py
```diff
@@ -72,14 +78,20 @@ class _NUFFT_ADJOP(torch.autograd.Function):
...
         return nufft_op.adj_op(y)
 
     @staticmethod
     def backward(ctx, dx):
         """Backward kspace -> image."""
-        (y, traj) = ctx.saved_tensors
+        if ctx.nufft_op._grad_wrt_data and not ctx.nufft_op._grad_wrt_traj:
+            y = ctx.saved_tensors
+        else:
+            (y, traj) = ctx.saved_tensors
```
paquiteau, line 94: same

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -745,6 +745,21 @@ def _safe_squeeze(self, arr):
...
+            ----------
+            new_smaps: C-ordered ndarray or a GPUArray.
+
+            """
+            if self.smaps_cached:
+                self.smaps = cp.array(
+                    new_smaps, order="C", copy=False, dtype=self.cpx_dtype
+                )
+            else:
+                self.smaps[...] = new_smaps.astype(self.cpx_dtype, copy=False)
```
paquiteau, line 762: why do you need the `[...]` ?

src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -502,6 +502,19 @@ def uses_sense(self):
...
+        Parameters
+        ----------
+        new_smaps: np.ndarray
+            the new sensitivity maps
+
+        """
+        smaps_ = new_smaps.T.reshape(-1, new_smaps.shape[0])
+        np.copyto(self.raw_op.pinned_smaps, smaps_)
+        self.smaps = new_smaps
+
```
paquiteau, line 517: this won't be enough for gpuNUFFT, you also need to make changes in the `RawGpuNUFFT` class, and do the proper management with respect to the smaps.

src/mrinufft/operators/autodiff.py
```diff
@@ -20,14 +20,20 @@ class _NUFFT_OP(torch.autograd.Function):
     @staticmethod
     def forward(ctx, x, traj, nufft_op):
         """Forward image -> k-space."""
-        ctx.save_for_backward(x, traj)
+        if nufft_op._grad_wrt_data and not nufft_op._grad_wrt_traj:
```
chaithyagr, line 23: This seems different, can you please document this update also in the PR ?

tests/operators/test_update.py
```diff
@@ -110,12 +110,22 @@ def kspace_data(operator):
...
     image_data = to_interface(image_data, array_interface)
     gitter = np.random.rand(*operator.samples.shape).astype(np.float32)
     # Add very little noise to the trajectory, variance of 1e-3
     operator.samples += gitter / 1000
+    if operator.smaps is not None and operator.backend in ['cufinufft', 'gpunufft']:
+        # Add noise to sensitivity maps
+        noise = np.random.rand(*operator.smaps.shape).astype(np.float32)
+        new_smaps = operator.smaps + noise
+        new_smaps /= np.linalg.norm(new_smaps, axis=0)
+        operator.update_smaps(new_smaps)
```
paquiteau, line 128: write a separate test for that.

tests/operators/test_update.py
```diff
@@ -133,6 +143,12 @@ def test_adj_op(
...
```
paquiteau, line 151: write a separate test for that

src/mrinufft/operators/interfaces/gpunufft.py
chaithyagr, line 517: I would rather recommend to have this as a property. Have a look at how samples.setter is dont just below. What we need is that the update of Smaps triggers all these compute internally. We dont want the user to do call this function.

paquiteau (changes requested):
Good starting ground! The `update_smaps` can be put in the `FourierOperatorBase` class (So that all backend have a default way of updating the smaps, for CPU Backends this is trivial). Also, I think we could be fine with a property setter (just as we did for the samples).

src/mrinufft/operators/interfaces/gpunufft.py
chaithyagr: Make sure to have the property defined in base class, with a setter function raising an error (just in case, if we have an interface which does not support smaps update). This setter can be overridden in respective downstream inherited classes later on.

chaithyagr (changes requested):
Great work, some minor comments and would prefer to have this as property rather than an additional function. It could be slightly more involved in code, but more cleaner surely.

src/mrinufft/operators/autodiff.py
chaithyagr: Good catch, i kept thinking that we are missing something surely, but moved focus of my review somewhere else.

src/mrinufft/operators/interfaces/gpunufft.py
chaithyagr: Umm why wont this be enough? The pointer to the pinned_smaps is pointed. As long as the data changes, it should be fine right?

src/mrinufft/operators/autodiff.py
paquiteau: I think thats for avoiding saving too many tensors. @AsmaTANABEN comments?

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -248,12 +248,12 @@ def __init__(
...
                     "used on gpu for smaps."
                 )
-                self.smaps = cp.array(
+                self._smaps = cp.array(
                     smaps, order="C", copy=False, dtype=self.cpx_dtype
                 )
                 self.smaps_cached = True
             else:
-                self.smaps = pin_memory(smaps.astype(self.cpx_dtype, copy=False))
+                self._smaps = pin_memory(smaps.astype(self.cpx_dtype, copy=False))
```
paquiteau, line 256: same story for the setter

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -264,6 +264,22 @@ def __init__(
...
+        new_smaps: C-ordered ndarray or a GPUArray.
+
+        """
+        if self.smaps_cached:
+            self._smaps = cp.array(
+                new_smaps, order="C", copy=False, dtype=self.cpx_dtype
+            )
+        else:
+            self._smaps = new_smaps.astype(self.cpx_dtype, copy=False)
+
```
paquiteau, line 282: Also check for the n_coils, and ideally image shape as well

tests/operators/test_update.py
```diff
@@ -63,29 +65,39 @@ def operator(
...
+        "samples": kspace_locs,
+        "shape": shape,
+        "n_coils": n_coils,
+        "n_batchs": n_batch,
+        "squeeze_dims": False,
+        "density": density,
+        "smaps": smaps,
+    }
+    if backend in ["cufinufft"]:
+        op_args.update([("smaps_cached", smaps_cached)])
```
paquiteau, line 78:
```suggestion
        op_args["smaps_cached"] = smaps_cached
```

tests/operators/test_update.py
```diff
@@ -63,29 +65,39 @@ def operator(
...
+        "samples": operator.samples,
+        "shape": operator.shape,
+        "n_coils": operator.n_coils,
+        "n_batchs": operator.n_batchs,
+        "squeeze_dims": False,
+        "density": operator.density,
+        "smaps": operator.smaps,
+    }
+    if operator.backend == "cufinufft":
+        op_args.update([("smaps_cached", operator.smaps_cached)])
```
paquiteau, line 97:
```suggestion
        op_args["smaps_cached"] = operator.smaps_cached
```

src/mrinufft/operators/autodiff.py
chaithyagr: Yes, i think, she needed this also, we can have it in same PR surely, but lets add that in description and/or title ?

tests/operators/test_update.py
```diff
@@ -1,4 +1,4 @@
-"""Test for update in trajectory and density.
+"""Test for update in trajectory,density, and sensitivity maps.
```
chaithyagr, line 1:
```suggestion
"""Test for update in trajectory, density, and sensitivity maps.
```

tests/operators/test_update.py
chaithyagr: I really thought this when I saw the code, but i thought its some "safe" way to do it perhaps xD

chaithyagr (approved):
Minor comments over what @paquiteau mentioned. But otheriwise good work! I think we can merge after you handling the shape checks

src/mrinufft/operators/autodiff.py
AsmaTANABEN: yeah it's about memory saving +  if differentiation is not required with respect to traj, it could be a non-tensor variable, which won't cause an error with the ctx.save_for_backward() function.

src/mrinufft/operators/base.py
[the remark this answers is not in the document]
AsmaTANABEN: Reverting to `self.smaps` will call the setter of the subclass. However, I need the behavior of the base setter in the constructor. To achieve this, I need to use `FourierOperatorBase.smaps.__set__(self, ...)`. any better suggestion ?

tests/operators/test_update.py
paquiteau: when you have a lot of update do to , this might be more efficient (I guess). But here we don't need that, just stay simple.

src/mrinufft/operators/autodiff.py
```diff
@@ -20,14 +20,15 @@ class _NUFFT_OP(torch.autograd.Function):
     @staticmethod
     def forward(ctx, x, traj, nufft_op):
         """Forward image -> k-space."""
-        ctx.save_for_backward(x, traj)
+        ctx.save_for_backward(x)
```
chaithyagr, line 23: @AsmaTANABEN just to let you know that I felt we dont need to savce the trajectory, so I removed that.

src/mrinufft/operators/autodiff.py
```diff
@@ -20,14 +20,15 @@ class _NUFFT_OP(torch.autograd.Function):
...
         ctx.nufft_op = nufft_op
+        ctx.dtype = traj.dtype
         return nufft_op.op(x)
 
     @staticmethod
     def backward(ctx, dy):
         """Backward image -> k-space."""
-        (x, traj) = ctx.saved_tensors
+        x = ctx.saved_tensors[0]
```
chaithyagr, line 31: Also, I fixed a minor bug in the code (which you would have n9oticed eventually)

chaithyagr (approved):
Added a few changes...

src/mrinufft/operators/autodiff.py
```diff
@@ -62,7 +63,7 @@ def backward(ctx, dy):
                     dim=0,
                 ),
                 dim=0,
-            ).type_as(traj)
+            ).to(ctx.dtype)
         return grad_data, grad_traj, None
```
paquiteau, line 67: you can use  `nufft_op.dtype` or `nufft_op.cpx_dtype`

tests/operators/test_update.py
```diff
@@ -63,29 +65,39 @@ def operator(
...
-    )
+    op_args = {
+        "samples": operator.samples,
+        "shape": operator.shape,
+        "n_coils": operator.n_coils,
+        "n_batchs": operator.n_batchs,
+        "squeeze_dims": False,
+        "density": operator.density,
+        "smaps": operator.smaps,
+    }
```
paquiteau, line 95:
```suggestion
   op_args = {k: getattr(operator, k) for k in ["samples", "shape", "n_coils", "n_batchs", "density", "smaps"]}
   op_args["squeeze_dims"] = False
```

tests/operators/test_update.py
```diff
@@ -138,3 +164,39 @@ def test_adj_op(
...
+
+
+@param_array_interface
+def test_op_smaps_update(
+    operator,
+    array_interface,
+    image_data,
+    new_smaps,
+):
+    """Test the batch type 2 (forward)."""
```
paquiteau, line 176:
```suggestion
    """Test the batch type 2 (forward) with smaps."""
```

tests/operators/test_update.py
```diff
@@ -138,3 +164,39 @@ def test_adj_op(
...
+
+
+@param_array_interface
+def test_adj_op_smaps_update(
+    operator,
+    array_interface,
+    kspace_data,
+    new_smaps,
+):
+    """Test the batch type 1 (adjoint)."""
```
paquiteau, line 194:
```suggestion
    """Test the batch type 1 (adjoint) with smaps."""
```

paquiteau (changes requested):
Some suggestion to improve further :) I am still worried about the update of _smaps directly, without check on shape, this could lead to bugs (or segfault in the backends)

src/mrinufft/operators/autodiff.py
chaithyagr: Good catch

chaithyagr:
>  I am still worried about the update of _smaps directly, without check on shape, this could lead to bugs (or segfault in the backends)

@AsmaTANABEN can you please handle this?

src/mrinufft/operators/autodiff.py
chaithyagr: Nope, reverting this :P  https://github.com/mind-inria/mri-nufft/actions/runs/10054335386/job/27788758132?pr=156
Turns out we need a torch.dtype and torch does not understand numpy dtype:

[1] https://github.com/pytorch/pytorch/issues/40568
[2] https://discuss.pytorch.org/t/converting-a-numpy-dtype-to-torch-dtype/52279

paquiteau: But we have NP2TORCH exactly for this (in `mrinufft._utils`) ;)

src/mrinufft/operators/base.py
```diff
@@ -500,15 +500,15 @@ def smaps(self):
...
     @smaps.setter
-    def smaps(self, smaps):
+    def smaps(self, smaps, check_only=False):
         if smaps is None:
             self._smaps = None
         elif len(smaps) != self.n_coils:
             raise ValueError(
                 f"Number of sensitivity maps ({len(smaps)})"
                 f"should be equal to n_coils ({self.n_coils})"
             )
```
paquiteau, line 510: I don't find any good reason to not also check that the remaining dimensions are matching self.shape as well. 

```suggestion
        elif smaps.shape != (self.n_coils, *self.shape):
            raise ValueError(
                f"smaps shape is {smaps.shape}, it should be"
                f"(n_coils, *shape): {(self.n_coils, *self.shape)}"
            )
```

src/mrinufft/operators/base.py
```diff
@@ -500,15 +500,15 @@ def smaps(self):
         return self._smaps
 
     @smaps.setter
-    def smaps(self, smaps):
```
paquiteau, line 503: I think this is quite complex, maybe refactor the dim check in a dedicated method (`self._check_smaps(smaps)`) that you call instead of using the fset attribute

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -221,35 +221,54 @@ def __init__(
...
+                if self._smaps is None:
+                    self._smaps = pin_memory(
+                        new_smaps.astype(self.cpx_dtype, copy=False)
+                    )
+                    self._smap_d = cp.empty(self.shape, dtype=self.cpx_dtype)
+                else:
+                    # copy the array to pinned memory
+                    np.copyto(self._smaps, new_smaps.astype(self.cpx_dtype, copy=False))
+        else:
+            self._smaps = new_smaps
```
paquiteau, line 272: Good Job

src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -202,6 +202,17 @@ def _reshape_image(self, image, direction="op"):
                 return image.squeeze().astype(xp.complex64, copy=False).T[None]
             return xp.asarray([c.T for c in image], dtype=xp.complex64).squeeze()
 
+    def set_smaps(self, smaps):
```
paquiteau, line 205: `set_pinned_smaps` ? But if you prefer to remain consistent with other backends, so be it.

paquiteau (changes requested):
Very minors comments, we will soon :tm: merge this

src/mrinufft/operators/interfaces/gpunufft.py
chaithyagr: I think its better to be consistent

src/mrinufft/operators/base.py
chaithyagr: Peace xD.

chaithyagr:
Disabling tensorflow for tests as it is painfully unstable and we now have seg faults due to some weird mismatch in driver. It is quite frustating. The main error was in converting cupy array to tensorflow tensor.

Issue #153, opened 2024-07-04
Get the alignment from the input data and pass it to header


chaithyagr:
Currently, we dont pull out this data. However, it is crucial information for us to finally save the reconstructed images. 

This could involve moving from pymapvbvd to twixtools if needed.

chaithyagr:
This isn't completed, we did a fix which didn't work so we reverted

PR #157, opened 2024-07-11
Twixtools and orientation matrix extraction

mbertrait:
This PR solves #153 by moving from [mapVBVD](https://github.com/wtclarke/pymapvbvd) to [twixtools](https://github.com/pehses/twixtools) to read Siemens Twix scan files and storing the orientation matrix in the `hdr` dict.

### DIFF
--- a/src/mrinufft/io/siemens.py
+++ b/src/mrinufft/io/siemens.py
@@ -5,10 +5,6 @@ from __future__ import annotations
 import numpy as np
 
 
-class TwixObj:
-    
-
-
 def read_siemens_rawdat(
     filename: str,
     removeOS: bool = False,
--- a/src/mrinufft/io/siemens.py
+++ b/src/mrinufft/io/siemens.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import numpy as np
 
+from .utils import get_siemens_twix_orientation_matrix
+
 
 def read_siemens_rawdat(
     filename: str,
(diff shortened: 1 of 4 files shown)


src/mrinufft/io/siemens.py
```diff
@@ -42,47 +44,55 @@ def read_siemens_rawdat(
...
-        "n_coils": int(twixObj.image.NCha),
-        "n_shots": int(twixObj.image.NLin),
-        "n_contrasts": int(twixObj.image.NSet),
-        "n_adc_samples": int(twixObj.image.NCol),
-        "n_slices": int(twixObj.image.NSli),
+        "n_coils": int(twixObj['image'].shape[-2]),
+        "n_shots": int(twixObj['image'].shape[-3]),
+        "n_contrasts": int(twixObj['image'].shape[-4]),
+        "n_adc_samples": int(twixObj['image'].shape[-1]),
+        "n_slices": int(int(twixObj['image'].shape[-5])),
```
paquiteau, line 71:
```suggestion
k: int(v) for k,v in zip(["n_adc_samples", "n_coils", "n_shots", "n_contrasts", "n_slices"], twixObj["image"].shape[::-1]) 
```

paquiteau (changes requested):
Nice ! I have made a suggestion for more pythonic code. Also you need to run black/ ruff on your code

src/mrinufft/io/siemens.py
mbertrait: I wanted to stick with the previous code "format" but sure

paquiteau: well at least put them in the correct order :D

chaithyagr (approved):
LGTM, I think Only Pierre's comments makes sense and some code linting. Ideally we need a test, but thats hard as we need to create dummy data. Lets skip testing this module but maybe @mbertrait can you open an issue if its not already present to track the need for a test?

chaithyagr:
A minor comment, if you can it could help to just tell here is twixtools is faster or pymapvbvd. It doesnt have to be an extensive test. Careful though, we dont want bad numbers just because data was in cache.

src/mrinufft/io/siemens.py
mbertrait: You're right maybe my goal of "sticking with the previous code and only change func calls etc." was at the cost of "less cleaner code". Let's go your way

chaithyagr (changes requested):
Please also change `pymapvbvd` to `twixtools` with the right github link in pyproject.toml
@paquiteau to comment how to specify gitghub link, as twixtools is not released by authors.

mbertrait:
> LGTM, I think Only Pierre's comments makes sense and some code linting. Ideally we need a test, but thats hard as we need to create dummy data. Lets skip testing this module but maybe @mbertrait can you open an issue if its not already present to track the need for a test?

I thought about tests but then as you mentioned we need dummy data so I set them aside for the moment.

mbertrait:
> Please also change `pymapvbvd` to `twixtools` with the right github link in pyproject.toml @paquiteau to comment how to specify gitghub link, as twixtools is not released by authors.

@paquiteau I found a syntax you can find in https://github.com/mind-inria/mri-nufft/pull/157/commits/612cfd9ce23613296376523138e89be10939a3a2 need your expertise on this

chaithyagr (approved):

paquiteau (approved):

PR #158, opened 2024-07-11
remove use_smap + fix pipe test

Lenoush:
- use compute_smaps instead of use_smap in the init of MRITensorflowNUFFT
- fix nm_iter into num_iterations in the pipe function 
- change syntaxe of pipe test because tensorflow mri  don't use osf (oversampling factor)

### DIFF
--- a/tests/operators/test_density_for_op.py
+++ b/tests/operators/test_density_for_op.py
@@ -1,6 +1,7 @@
 """Specific test for testing densities specific to backend."""
 
 import numpy as np
+import pytest
 from pytest_cases import parametrize, parametrize_with_cases
 
 from case_trajectories import CasesTrajectories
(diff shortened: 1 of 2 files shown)


tests/operators/test_density_for_op.py
```diff
@@ -27,8 +28,14 @@ def radial_distance(traj, shape):
 @parametrize(backend=["gpunufft", "tensorflow"])
 def test_pipe(backend, traj, shape, osf):
     """Test the pipe method."""
+    if backend == "tensorflow" and osf == 1:
+        result = pipe(traj, shape, backend, num_iterations=10)
+    elif backend == "tensorflow" and osf != 1:
+        pytest.skip("Tensorflow does not support OSF != 1")
+    else:
+        result = pipe(traj, shape, backend, osf=osf, num_iterations=10)
```
paquiteau, line 36:
```suggestion
    if backend == "tensorflow" and osf != 1:
        pytest.skip("Tensorflow does not support OSF != 1")
    result = pipe(traj, shape, backend, osf=osf, num_iterations=10)
```

paquiteau: just simplifying a bit.

paquiteau (changes requested):
Nice, do you wish to have this run in the CI as well ?

tests/operators/test_density_for_op.py
Lenoush: i thought of doing this but tensorflow mri doesn't take osf arguments: so to avoid having an unused argument in pipe i added conditions in test

Lenoush: we can add this in the CI , i test on my machine and it work

paquiteau: > 

i thought of doing this but tensorflow mri doesn't take osf arguments: so to avoid having an unused argument in pipe i added conditions in test

In this case, I would recommend having osf=1 as a dummy argument in tfnufft pipe (and raise an error there). This would make a consistent API call for all pipe methods.

chaithyagr: By default osf=2 internally within `tensorflow-mri` I recommend using the same default value.

chaithyagr:
I would wait for #160 to merge to master before merging this PR. We can see if tests are green then.

chaithyagr:
Closing as this is now taken in #160, which also does the tests for tensorflow-nufft.

PR #160, opened 2024-07-12
Plugging a bunch of test holes back to CI

### LINKED ISSUE #140: Fix a test hole for ref_backend, i.e. `finufft`
Currnetly in:

https://github.com/mind-inria/mri-nufft/blob/71f1cf0c781210b35907e8686a9b4b8dfc522b98/tests/conftest.py#L84-L87

This was added originally to ensure we dont blow up the number of tests, however, we seem to be missing some basic tests with `finufft` like `test_batch` and `test_autodiff` which was seen in #139 .

In #139, only closing this, the issue turned to be more complicated, with a lot of failures (some handeled in #139) and also a lot of extra tests running (for example `finufft` tests in `cufinufft`.

To fix this, we might need to rediscuss our strategy on how to tackle it.

chaithyagr:
It seems like tensorflow NUFFT and also let us resolve #140  in this PR


chaithyagr:
I merged #158 here to take the changes to fix tensorflow-nufft. This way we can have passing tests as we merge.

tests/conftest.py
```diff
@@ -81,8 +81,10 @@ def pytest_generate_tests(metafunc):
...
-            if backend == metafunc.config.getoption("ref"):
-                # FIXME: This is skipping most tests on finufft, #140
+            if metafunc.config.getoption(
+                "backend"
+            ) != [] and backend != metafunc.config.getoption("backend"):
+                # Skip tests if the backend does not match what we want to test.
                 callspec.marks.append(
-                    pytest.mark.skip("Not testing ref backend with self.")
+                    pytest.mark.skip(f"Not testing {backend} as not requested")
                 )
```
chaithyagr, line 90: This is slightly more sophisticated, as we moved from skipping if we are testing with reference to
testing only if we request for it. Else, we skip. @paquiteau to comment in case this causes new test holes.

src/mrinufft/operators/interfaces/gpunufft.py
```diff
@@ -511,12 +511,14 @@ def samples(self, samples):
         samples: np.ndarray
             The samples for the Fourier Operator.
         """
+        self._samples = samples
```
chaithyagr, line 514: This was a bug we missed earlier. We used to compute density without updating samples. Fixed this now. However, I added a TODO that we still have performance issue, this needs gpuNUFFT to sort the samples twice, once for pipe and once for the operator. While we do have `pipe` method. It currently is messy to just call it directly (as we have `osf` and `upsampfac` as 2 different parameters. Also, we still create an operator to ensure we can call pipe without user needing to create a new fourier_op. @paquiteau whats the best way forward here? Lets discuss this sometime as we have all the elements to get rid of perf issues but we need to agree on standard to follow. This can greatly benefit in learning trajectory

https://github.com/mind-inria/mri-nufft/blob/20a77aaeee0a2ac71002c92d99d59a8650ba474d/src/mrinufft/operators/interfaces/gpunufft.py#L551-L557

tests/conftest.py
paquiteau: I think this change is redundant with what is in line 20-30 above (we modify the availability of the backend dynamically in the tests config). I would rather advise to remove these lines (84-90) and see what happens.

The whole idea of this was to not test the backend against itself (it does not make sense, and it save time, esp. for the `test_interfaces.py` tests).

chaithyagr: Thats weird, in that case, for some reason we trigger autodiff tests for `finufft` when the requested backend is `tensorflow` :P . This is why I added it in first place..

src/mrinufft/operators/base.py
```diff
@@ -753,7 +781,7 @@ def _grad_sense(self, image_data, obs_data):
 
         dataf = image_data.reshape((B, *XYZ))
         obs_dataf = obs_data.reshape((B * C, K))
-        grad = np.empty_like(dataf)
+        grad = np.zeros_like(dataf)
```
chaithyagr, line 784: This was such a hard catch :P We missed this as the test assert_correlate would pass if we have `nan`

tests/operators/test_batch.py
```diff
@@ -186,16 +186,15 @@ def test_data_consistency(
...
     if len(operator.shape) == 2 and operator.uses_sense:
         print("Reduced accuracy for 2D Sense")
-        slope_err = 1e-1
-        r_value_err = 1e-1
+        atol = 1e-1
+        atol = 1e-1
 
-    for i in range(len(res)):
-        assert_correlate(res[i], res2[i], slope_err=slope_err, r_value_err=r_value_err)
+    npt.assert_allclose(res, res2, atol=atol, rtol=rtol)
```
chaithyagr, line 197: Moved assert_correlate to assert_allclose. @paquiteau to just confirm if we could be missing any tests due to assert_correlate which can be changed in this PR itself.

src/mrinufft/operators/base.py
paquiteau: nice catch !

chaithyagr:
76.96 to 77.89% in coverage ! Yayy xD

paquiteau (changes requested):

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -58,56 +58,55 @@ def __init__(
         **kwargs,
     ):
         self.shape = shape
-        self.samples = samples
         self.ndim = len(shape)
```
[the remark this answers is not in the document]
chaithyagr: We discussed it on board :) 
For posterity:

We dont want to have multiple states of samples (one in wrapper, one in raw_op and one inside raw_op). This can cause weird issues and make it hard to track

tests/helpers/factories.py
```diff
@@ -97,7 +97,12 @@ def wrapper(operator, array_interface, *args, **kwargs):
         else:
             op = operator
         if array_interface in ["torch-gpu", "cupy"]:
-            if op.backend not in ["cufinufft", "gpunufft", "torchkbnufft-gpu"]:
+            if op.backend not in [
+                "cufinufft",
+                "gpunufft",
+                "torchkbnufft-gpu",
+                "tensorflow",
```
paquiteau, line 104: do we support torch array with tensorflow backend ? Seems weird to me

paquiteau (approved):
LGTM !

tests/helpers/factories.py
chaithyagr: Yup xD, cause why not ... But yes, its not efficient at allll. There weirdly is no way to convert torch tensor to tensorflow tensor, without copying to CPU first... However, its just a way to prevent issues and bugs, it reallly isnt a usecase.

paquiteau: well, okay, I'll let you merge :)

Issue #150, opened 2024-07-01
Ensure that input image data is of right shape and order


chaithyagr:
We currently dont check if the image shape matches what was specified in the init.
Additionally, it will help a lot to ensure that we are C ordered in the data.

PR #161, opened 2024-07-12
150 right shape

Lenoush:
answer to https://github.com/mind-inria/mri-nufft/issues/150 

Basically, every time you call op and adj_op, make sure the input image is of the shape <XYZ>, which was provided during initialization.

- create a function that checks whether the shape of the image (in this case, the data) == self.shape (from init) 
- call it when op 
- test this function
- call it when adj_op

The CPU-based backend is handled differently


src/mrinufft/operators/base.py
```diff
@@ -531,6 +531,14 @@ def norm_factor(self):
...
         return np.sqrt(np.prod(self.shape) * (2 ** len(self.shape)))
 
+    def check_shape(self, image, init_shape):
+        """Check if the image shape is compatible with the operator's init shape."""
+        if image.shape != init_shape:
+            raise ValueError(
+                f"Image shape {image.shape} is not compatible with the operator shape"
+                f" {self.shape}"
+            )
+
```
chaithyagr, line 541:
```suggestion
    def check_shape(self, image):
        """Check if the image shape is compatible with the operator's init shape."""
        if image.shape[-len(self.shape):] != self.shape:
            raise ValueError(
                f"Image shape {image.shape} is not compatible with the operator shape"
                f" {self.shape}"
            )

```

src/mrinufft/operators/interfaces/torchkbnufft.py
```diff
@@ -123,6 +123,7 @@ def op(self, data, out=None):
         -------
         Tensor: Non-uniform Fourier transform of the input image.
         """
+        self.check_shape(self._safe_squeeze(data), self.shape)
```
chaithyagr, line 126:
```suggestion
        self.check_shape(self._safe_squeeze(data))
```

chaithyagr (changes requested):
I think we dont need to send the shape in. Also, we might need to be careful and check only last 2 or 3 dimensions (in case we have batches)

src/mrinufft/operators/base.py
Lenoush: Yes, you're right, that's why the test failed.

paquiteau (changes requested):
All CPU based backend should rather be handled via the `FourierOperatorCPU.op`. 

Otherwise this LGTM

tests/test_shape.py
```diff
@@ -0,0 +1,45 @@
...
+    [
+        ((256, 256), (16, 10, 256, 254)),
+        ((256, 256, 176), (16, 10, 256, 256)),
+    ],
+)
+def test_check_shape_fail(self_shape, image_shape):
+    image = np.random.rand(*image_shape)
+    expected_message = f"Image shape {image.shape[-len(self_shape):]} is not compatible with the operator shape {self_shape}"
+    with pytest.raises(ValueError, match=re.escape(expected_message)):
+        check_shape(self_shape, image)
```
chaithyagr, line 45: I dont think we need this test... It seems redundant...

chaithyagr (changes requested):

tests/test_shape.py
paquiteau: redundancy is not a bad things for tests. However, the testes `check_shape` must be the one in the utils module (otherwise this has no interest!)

chaithyagr: Oh come on, its literally same code in test xD

Lenoush: @paquiteau i know,but strangely it cannot import the function , it's was the just a first draft to see if i need to change something , i'm working on that

src/mrinufft/operators/interfaces/utils/utils.py
paquiteau, line 1: Is there any good reason not to make those methods of the `FourierOperatorBase` class ?

paquiteau (commented):
I think the check_shape functions could be methods of `FourierOperatorBase`, is there any good reason not to do that ? @chaithyagr

src/mrinufft/operators/interfaces/utils/utils.py
chaithyagr: No, i thought it was in base class. It should be in base class indeed.

src/mrinufft/operators/base.py
```diff
@@ -275,6 +278,55 @@ def __init_subclass__(cls):
         if backend := getattr(cls, "backend", None):
             cls.interfaces[backend] = (available, cls)
 
+    def check_shape_op(self, image, ksp=None):
```
chaithyagr, line 281: Why give ksp when you dont test its shape anyways?  We need to ideally check kspace's shape here if it is not None. I would recommend having the same function for op and adj_op

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -299,11 +299,7 @@ def op(self, data, ksp_d=None):
         this performs for every coil \ell:
         ..math:: \mathcal{F}\mathcal{S}_\ell x
         """
-        # monocoil
-        if self.uses_sense:
-            check_size(data, (self.n_batchs, *self.shape))
-        else:
-            check_size(data, (self.n_batchs, self.n_coils, *self.shape))
+        self.check_shape_op(self, data)
```
chaithyagr, line 302: That way here you send `data, coeff`

src/mrinufft/operators/base.py
```diff
@@ -275,6 +278,55 @@ def __init_subclass__(cls):
...
+
+        Returns
+        -------
+        None
+            This function does not return any value. It raises a ValueError if the
+            image shape does not match the expected shape.
+        """
+        image_samples = (
+            image.n_samples if hasattr(image, "n_samples") else image.shape[-1]
+        )
```
chaithyagr, line 322: I dont know if it is image. Why are we comparing image.n_samples here? shouldnt we just check image.shape?

chaithyagr: I think you meant kspace?

chaithyagr (changes requested):
Thank you. I think the codes are a bit confusing and some things can be refactored

src/mrinufft/operators/base.py
```diff
@@ -275,6 +278,41 @@ def __init_subclass__(cls):
...
             cls.interfaces[backend] = (available, cls)
 
+    def check_shape(self, image=None, ksp=None):
+        """
+        Validate the shapes of the image or k-space data against operator shapes.
+
+        Parameters
+        ----------
+        image : np.ndarray, optional
+            In "op" function, the image data will be checked.
```
chaithyagr, line 288:
```suggestion
            If passed, the shape of image data will be checked.
```

src/mrinufft/operators/base.py
```diff
@@ -275,6 +278,41 @@ def __init_subclass__(cls):
...
+
+        ksp : np.ndarray or object, optional
+            In "adj_op" function, the k-space data will be checked.
```
chaithyagr, line 291:
```suggestion
            If passed, the shape of the k-space data will be checked.
```

src/mrinufft/operators/base.py
```diff
@@ -711,6 +750,8 @@ def adj_op(self, coeffs, img=None):
         -------
         Array in the same memory space of coeffs. (ie on cpu or gpu Memory).
         """
+        self.check_shape(image=img, ksp=coeffs)
+
```
chaithyagr, line 754:
```suggestion
```

src/mrinufft/operators/interfaces/bart.py
```diff
@@ -142,6 +142,8 @@ def __init__(
             else:
                 kwargs["extra_adj_op_args"] = ["-i"]
 
+        self.raw_op = RawBartNUFFT(samples_, shape, **kwargs)
+
```
chaithyagr, line 146:
```suggestion
```

tests/test_shape.py
```diff
@@ -0,0 +1,100 @@
...
+    from_interface,
+    param_array_interface,
+)
+
+
+@fixture(scope="module")
+@parametrize(
+    "backend",
+    [
+        "torchkbnufft-gpu",
```
chaithyagr, line 26: Why only torchkbnufft-gpu? Please add others also if it was by mistake?

chaithyagr (changes requested):
Almost there, getting nitpicky now, but a couple of important changes also.

tests/test_shape.py
Lenoush: I just put one at first and forgot to add others ones :)

paquiteau (changes requested):
Some suggestions, otherwise this seems good.

src/mrinufft/operators/base.py
```diff
@@ -275,6 +278,41 @@ def __init_subclass__(cls):
         if backend := getattr(cls, "backend", None):
             cls.interfaces[backend] = (available, cls)
 
+    def check_shape(self, image=None, ksp=None):
```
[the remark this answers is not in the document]
Lenoush: I did'nt even know this exist ! great !!

PR #162, opened 2024-07-15
Fix twixObj for reading the orientation matrix right.

chaithyagr:
There was a bug in earlier codes, fixing them here. This is probably why a test can be useful :P 
Adding @mbertrait to actually review and run the codes to ensure we dont have bugs,.

### DIFF
--- a/src/mrinufft/io/siemens.py
+++ b/src/mrinufft/io/siemens.py
@@ -59,21 +59,22 @@ def read_siemens_rawdat(
             "The twixtools module is not available. Please install it using "
             "the following command: pip install gt-twixtools"
         ) from err
-    twixObj = map_twix(read_twix(filename))
+    twixObj = read_twix(filename)
     if isinstance(twixObj, list):
         twixObj = twixObj[-1]
-    twixObj["image"].flags["remove_os"] = removeOS
+    twix = map_twix(twixObj)
+    twix["image"].flags["remove_os"] = removeOS
     hdr = {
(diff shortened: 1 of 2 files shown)


chaithyagr:
Still I see issues while reading Siemens data:

```

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[46], line 3
      1 cart_levels = []
      2 for i, file in enumerate(['meas_MID00349_FID45226_3DT1.dat', 'meas_MID00352_FID45229_3DT1_0V.dat']):
----> 3     cart_levels.append(read_siemens_rawdat(os.path.join(folder, 'raw_data', file)))

File /volatile/Chaithya/Environments/nufft/lib/python3.10/site-packages/mrinufft/io/siemens.py:77, in read_siemens_rawdat(filename, removeOS, squeeze, return_twix, slice_num, contrast_num)
     75 hdr["shifts"] = ()
     76 for s in [7, 6, 8]:
---> 77     shift = twix["hdr"]["Phoenix"]["sWipMemBlock"]["adFree"][s]
     78     hdr["shifts"] += (0,) if shift == [] else (shift,)
     80 hdr["orientation_matrix"] = get_siemens_twix_orientation_matrix(twixObj)

KeyError: 'sWipMemBlock'
```

It seems like we need to pick up the shifts from somewhere else, not rely on sWipMemBlock, which is specific to SPARKLING (arbgrad module)

PR #164, opened 2024-07-17
Fix init tf

Lenoush:
while doing the tensorflow benchmark, I realized that there were still some problems with the smaps. 
It should be fine now :)


chaithyagr:
I am not sure if everything is covered. @Lenoush please rebase your changes to #160 and then see if you still had changes.

paquiteau:
No need for this, we checked everything with @Lenoush

Issue #170, opened 2024-07-29
Prevent saving temporary images on `/tmp` in examples


chaithyagr:
Saving temporarily on `/tmp` causes issues on local runners as some tasks seem to be deleting the files even before it is accessed, causing weird errors. @paquiteau I wonder if this is because the `shasum` is matching that from other examples?

paquiteau:
that is very likely indeed, a solution would be to add some salt on the hash, aka a UNIX timestamp.

PR #171, opened 2024-07-29
fix: add unix time stand for unicity.

paquiteau:
Solves #170 ?


chaithyagr (changes requested):
I'm wondering if we should in any case just create the images within the same folder? I'm worried if the time stamp is not enough for unicity, as all could have same time when we run with parallel

paquiteau:
honestly, having the same timestamp, down to the tenth of microsecond (`time.time()` is a float !) should be rare enough not to happen. It is a risk I am willing to take :D otherwise, we could a swell add os pid in the hash but that may be just overkill

chaithyagr (approved):
LGTM.. For now this is not done in the learning example, but I think thats okay for now.

PR #173, opened 2024-08-01
Some tweaks for stacked nufft

paquiteau:
- make z_index="auto" default 
- renable test for stacked-gpu.


src/mrinufft/operators/stacked.py
```diff
@@ -74,9 +74,9 @@ def __init__(
         self,
         samples,
         shape,
-        z_index,
         backend,
         smaps,
+        z_index="auto",
```
chaithyagr, line 79: Perhaps update the docs also for it?

chaithyagr (changes requested):
A minor catch.. You can merge after resolving it and after getting green

PR #174, opened 2024-08-05
Fix MaxRecursionDepth error, Memory leaks and move to CUDA12

chaithyagr:
We had some issues of MaxRecurssionDepth, which was fixed and on extensively testing, I saw a memory leak in gpuNUFFT (fixed in https://github.com/chaithyagr/gpuNUFFT/pull/23). Some more memory leaks from cuFFT comes from from CUDA11.8: https://forums.developer.nvidia.com/t/nvc-cufft-memory-leak/279476/3

To fix this, we moved to using CUDA12.1, which is default version which comes with `torch`. 

Also have a bunch of other miscellaneous tasks:

1) Added support for binder button in docs.
2) Fixed a minor bug in smaps estimation which didnt allow `int` values for `blurr_factor`.


.github/workflows/test-ci.yml
```diff
@@ -282,7 +280,7 @@ jobs:
   BuildDocs:
     name: Build API Documentation
     runs-on: gpu
-    if: ${{ (contains(github.event.head_commit.message, 'docs_build')) || (github.ref == 'refs/heads/master')}}
+    if: ${{ !(contains(github.event.head_commit.message, 'docs_build')) || (github.ref == 'refs/heads/master') }}
```
paquiteau, line 283: so whats the expected behavior ? one needs to put `docs_build` to skip the building of the documentation ? seems weird to me...

src/mrinufft/extras/smaps.py
```diff
@@ -171,7 +171,7 @@ def low_frequency(
         Smaps = Smaps * convex_hull
     # Smooth out the sensitivity maps
     if np.sum(blurr_factor) > 0:
-        if isinstance(blurr_factor, float):
+        if isinstance(blurr_factor, float) or isinstance(blurr_factor, int):
```
paquiteau, line 174:
```suggestion
        if isinstance(blurr_factor, (float, int)):
```

src/mrinufft/operators/base.py
```diff
@@ -423,11 +423,12 @@ def compute_density(self, method=None):
...
-            samples,
-            shape,
-            **kwargs,
-        )
+        if self.density_method is None:
+            self.density_method = lambda samples, shape: method(
+                samples,
+                shape,
+                **kwargs,
+            )
```
paquiteau, line 431: I think it would be best to make _density_method a private attribute, the user should not have to interact with it.

paquiteau (changes requested):
Great Work ! moving to cuda12 is really nice. We should probably make a release after this.

I have some minor comments/suggestions to address before that though ;)

.github/workflows/test-ci.yml
chaithyagr: Its weird, when I added this, it is building .. I am lost but it works :P

paquiteau: Let's clear this then, We should build the doc: 
- In PR to master, when the `docs_build` keyword is present.
- On every commit to master (aka after merge)

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#github-context 

I think we should use ``github.head_ref`` (and the contains)

chaithyagr: The ref aspect is fine. The containts part doesnt work. I am trying to understand whats broken. For some reason it does not trigger even when I use `docs_build`

chaithyagr: Okay this is broken. For now, we dont check `docs_build` and also `style`. (Cause `github.event.head_commit` is empty if the event is triggered with `pull_request` I will open an issue to fix this, but its not easy :P 
We can proceed with current PR for now,.

paquiteau (approved):

Issue #124, opened 2024-06-04
Add quick utils functions to get B and C matrices from off-resonance maps


chaithyagr:
Currently, in

https://github.com/mind-inria/mri-nufft/blob/f5a6ba006184b96a1f73ff54be27d854797c79fe/src/mrinufft/operators/off_resonnance.py#L19-L58

we just generate the fourier_op which takes the B and C values into account. 

- [ ] It might help to also include a function in `mrinufft.operators.off_resonance` called `get_interpolators_from_fieldmap`.
- [ ] Also, it might further help to be able to get fourier correction possible directly from the __init__ function itself like `MRIFourierCorrected(fourier_op, b0_field, mask)`
- [ ] It could be useful to allow for autodiff to pass through this new fourier op, currently we only support cupy or numpy, it might help to check if we have autograd enabled on based fourier_op and enable it as well (we just need to have tensors as torch tensors.
- [ ] A bit distant, but it could help to have the off-resonance estimation from k-space data as function. Although, this is a bit distant and I really feel it should belong to pysap-mri.

paquiteau:
I fully agree with you. 

Getting the off resonance map from the k space data could be done in the extras module, just as we propose it  for the smaps estimation. 

We may also wish to  have autodiff support around this as well ?

chaithyagr:
Sure. But note that the estimation isnt trivial, it is computationally expensive (for now). @Daval-G you think these codes can be open sourced though? I wanted to confirm that.

philouc:
I agree with @paquiteau : Estimation of off-resonance effects from k-space data should be carried out separately, as it depends on the phase information and TE value usually extracted from the image domain.

PR #180, opened 2024-08-09
Enhance Fourier Corrected operator

mcencini:
Hi all,
thanks again for this package! This PR partially address #124.

Specifically, I added a `get_interpolators_from_fieldmap` (supporting complex R2* + 1i * B0 fields) and modified the MRIFourierCorrected constructor to use this.

I also added an example for Fourier Corrected operator (I think this was missing).
All the best,
Matteo


paquiteau (commented):
Hi (again !) Thanks for this PR this is long due in MRI-NUFFT.

From a brief reading, this seems well done; I have some advice for the tests, though.

I will let @chaithyagr and @Daval-G review the math part. We  can probably use this PR as well to add the ORC code of @Daval-G [^0](which uses the SVD approach[^1], and, if I recall correctly, brought the best performances). Your code is GPU-capable, so that's a strong advantage nonetheless.

PS: Most of the team is taking vacation in the coming weeks, so their review will likely be delayed.

[^0]: https://github.com/CEA-COSMIC/pysap-mri/blob/develop/src/mri/operators/fourier/utils/orc_coefficients.py 
[^1]: Sutton, Noll & Fessler (2003-02) Fast, Iterative Image Reconstruction for MRI in the Presence of Field Inhomogeneities, IEEE Transactions on Medical Imaging.

mcencini:
Thanks a lot for the feedback! I will be on vacation in the next weeks as well and go back to this when I return :D

.github/workflows/test-ci.yml
```diff
@@ -189,7 +189,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -e .[test,dev]
-          python -m pip install finufft pooch brainweb-dl torch
+          python -m pip install finufft pooch brainweb-dl torch phantominator
```
chaithyagr, line 192: I think we can also stick to brainweb-dl? Is there any reason why you need to use phantominator?

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+
+# %%
+# Data Generation
+# ===============
+# As example, 2D shepp-logan will be used.
+# installable using ``pip install phantominator``
+
+from phantominator import ct_shepp_logan
+
+mri_data = ct_shepp_logan(256)
```
chaithyagr, line 29: You can have a look at other examples where a slice of 2D MRI is chosen from brainweb-dl.

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+    rr = (xx**2 + yy**2) ** 0.5
+
+    # mask
+    mask = (obj != 0).astype(np.float32)
+    b0map = mask * rr
+
+    # rescale
+    b0map = (b0range[1] - b0range[0]) * b0map / b0map.max() + b0range[0]  # Hz
+
+    return b0map * mask
```
chaithyagr, line 70: This can be a function / fixture can can be used in tests too!

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+
+# %%
+# Generate a Spiral trajectory
+# ----------------------------
+
+from mrinufft import initialize_2D_spiral
+from mrinufft.density import voronoi
+
+samples = initialize_2D_spiral(Nc=16, Ns=2000, nb_revolutions=10)
+t_read = np.arange(samples.shape[1]) * 4e-6  # 1us raster time
```
chaithyagr, line 84: We could use `DEFAULT_RASTER_TIME` from: https://github.com/mind-inria/mri-nufft/blob/aad8c0e77f6c059d139e2ba37ce7f78e0914a787/src/mrinufft/trajectories/utils.py#L15

This will help a lot to keep things more uniform. Also, I am curious if you meant 4us raster time in the comments?

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+import mrinufft
+from mrinufft.operators.off_resonance import MRIFourierCorrected
+
+# Generate standard NUFFT operator
+NufftOperator = mrinufft.get_operator("finufft")
+nufft = NufftOperator(
+    samples=samples,
+    shape=mri_data.shape,
+    density=density,
+)
```
chaithyagr, line 103:
```suggestion
nufft = mrinufft.get_operator("finufft")(
    samples=samples,
    shape=mri_data.shape,
    density=density,
)
```


Also, we can just import `get_operator` above to simplify the code further I think

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Generate Fourier Corrected operator
+mfi_nufft = MRIFourierCorrected(nufft, fieldmap=b0map, t=t_read, mask=mri_data != 0)
```
chaithyagr, line 106: The mask part may not work always (if data is noisy for example). I think it could be helpful to have a mask estimator code from @Daval-G present somewhere in utils and piping that here.

chaithyagr: Perhaps a more robust estimator can do a mean:
```suggestion
mfi_nufft = MRIFourierCorrected(nufft, fieldmap=b0map, t=t_read, mask=mri_data >=0.5*np.mean(mri_data))
```

examples/example_offresonance.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Reconstruct without field correction
+mri_data_adj = nufft.adj_op(kspace)
+mri_data_adj = np.squeeze(abs(mri_data_adj))
+print(mri_data_adj.shape)
+
+# Reconstruct with field correction
+mri_data_adj_mfi = mfi_nufft.adj_op(kspace)
+mri_data_adj_mfi = np.squeeze(abs(mri_data_adj_mfi))
+print(mri_data_adj_mfi.shape)
```
chaithyagr, line 120: I think we can remove the print statements? I am assuming you added them for easier debug, but they are not needed in final example I feel

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,325 @@
...
+    import cupy as cp
+except ImportError:
+    CUPY_AVAILABLE = False
+
+
+TORCH_AVAILABLE = True
+try:
+    import torch
+except ImportError:
+    TORCH_AVAILABLE = False
```
chaithyagr, line 26: I think you can pull `CUPY_AVAILABLE` and `TORCH_AVAILABLE` from `.base`:
https://github.com/mind-inria/mri-nufft/blob/aad8c0e77f6c059d139e2ba37ce7f78e0914a787/src/mrinufft/operators/base.py#L22-L34

 (Note that TORCH_AVAILABLE is called `AUTOGRAD_AVAILBLE`

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,325 @@
...
+
+    Here, B has shape ``(n_time_segments, len(t))``
+    and C has shape ``(n_time_segments, *fieldmap.shape)``.
+
+    From Sigpy: https://github.com/mikgroup/sigpy
+    and MIRT (mri_exp_approx.m): https://web.eecs.umich.edu/~fessler/code/
+
+    Parameters
+    ----------
+    fieldmap : np.ndarray or GPUarray
```
chaithyagr, line 42: Do we use GPUArray @paquiteau. Just want to ensure documentation is streamlined.

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,325 @@
...
+    # back to torch if required
+    if is_torch:
+        if xp.__name__ == "cupy":
+            B = torch.from_dlpack(B)
+            C = torch.from_dlpack(C)
+        else:
+            B = torch.from_numpy(B)
+            C = torch.from_numpy(C)
+
+    return B, C
```
chaithyagr, line 169: I think we have `with_numpy_cupy` function which internally converts everything to numpy or cupy. Feel free to use it to prevent redundancy:

https://github.com/mind-inria/mri-nufft/blob/aad8c0e77f6c059d139e2ba37ce7f78e0914a787/src/mrinufft/operators/base.py#L178-L216

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,325 @@
...
+    backend: str, optional
+        The backend to use for computations. Either 'cpu', 'gpu' or 'torch'.
+        The default is `cpu`.
+    """
+
+    def __init__(
+        self,
+        fourier_op,
+        fieldmap=None,
+        t=None,
```
chaithyagr, line 227: Consider calling it `time` for better understanding.

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,325 @@
...
+            N-D input image
+        obs_data: numpy.ndarray or cupy.ndarray
+            N-D observed k-space
+
+        Returns
+        -------
+        numpy.ndarray or cupy.ndarray
+            data consistency error in image space.
+        """
+        return self.adj_op(self.op(image_data) - obs_data)
```
chaithyagr, line 325: We call this `data_consistency` and we already have it defined, this is not needed.
```suggestion
```

tests/operators/test_offres_exp_approx.py
```diff
@@ -0,0 +1,241 @@
...
+
+def make_b0map(ndim, npix=64):
+    """Make B0 map (units: Hz)."""
+    # generate support
+    shape = ndim * [npix]
+    if ndim == 2:
+        mask, fieldmap = _make_disk(shape)
+    elif ndim == 3:
+        mask, fieldmap = _make_sphere(shape)
+
```
chaithyagr, line 68: I like these functions, it could help to have it somewhere internally supported. Also @paquiteau it could help to add your complex B0 map maker from brainweb-dl somewhere (maybe it belongs in brainweb-dl, not sure..)

tests/operators/test_offres_exp_approx.py
```diff
@@ -0,0 +1,241 @@
...
+    return rr < frac_radius, rr
+
+
+def get_backend(backend):
+    if backend == "cpu":
+        return np
+    elif backend == "gpu":
+        return cp
+    elif backend == "torch":
+        return torch
```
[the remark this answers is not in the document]
chaithyagr: You mean reinvent what you already re-invented? :P

chaithyagr (changes requested):
Thank you for this wonderful contribution! A bunch of comments and questions from me is below. Also, we need a docs_build for this to ensure documentation builds fine. @paquiteau I think thats on you and me to handle though.

Also, I will let @Daval-G handle more specific comments on B0 parts of the codes..

tests/operators/test_offres_exp_approx.py
paquiteau: I don't have support (yet) for B0 map in `brainweb-dl` (but 3 maps are available for download https://brainweb.bic.mni.mcgill.ca/about_sbd.html). Generating trivial B0-maps could be setup in `mrinufft.extras` (along with some smaps generation code I keep copy pasting everywhere)

src/mrinufft/operators/off_resonance.py
mcencini: I think this was already in the code (perhaps some initial version before data_consistency was defined?). I agree this is redundant now!

src/mrinufft/operators/off_resonance.py
mcencini: I tried before, but the expected signature for "with_numpy_cupy" (data, output=None, *args, **kwargs) seems to not work well with get_interpolators_from_fieldmap (while I guess it is generally ok for nufft routines).

src/mrinufft/operators/off_resonance.py
mcencini: Do you think `readout_time` would be too verbose?

examples/example_offresonance.py
mcencini: My intention was to prepare a simple example with the sole purpose of showing the effect of field inhomogeneities. For a real noisy dataset something like `>=0.5*np.mean(mri_data)` would probably be suboptimal anyway (holes, spurious points in the background). Maybe it would be better to explain this in the script text? Anyway, I am happy to replace the line as you suggested if you think this is better :)

examples/example_offresonance.py
mcencini: I am not really sure how to avoid copying the code here and in tests without incorporating these in the mrinufft._utils. Do you have any suggestion?

.github/workflows/test-ci.yml
mcencini: I was having some SSL verification problem and decided to use this. I am trying to solve this.

examples/example_offresonance.py
mcencini: Done, thanks!

examples/example_offresonance.py
mcencini: I agree! FYI, I adapted this from the example_stacked.py, which has the print statements. Shall we remove it from that example as well?

src/mrinufft/operators/off_resonance.py
mcencini: What about using npt.ArrayLike?

examples/example_offresonance.py
chaithyagr: Ouch, i didn't even know that example existed. I'll let Pierre Antoine comment on that and this then...

tests/operators/test_offres_exp_approx.py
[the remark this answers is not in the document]
mcencini: Thanks! Apparently this is not needed anymore after test refactor.

src/mrinufft/operators/off_resonance.py
chaithyagr: No I think for parameters it's ok to be verbose

tests/operators/test_offres_exp_approx.py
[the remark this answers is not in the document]
mcencini: I am not super familiar with fixtures, I hope I got that right.

examples/example_offresonance.py
paquiteau: Maybe in a `mrinufft/extras/data.py` file ?

src/mrinufft/operators/off_resonance.py
paquiteau: I would say that `ndarray` is good enough for a docstring, with a comment saying that this compatible with cupy and torch.

src/mrinufft/operators/off_resonance.py
paquiteau: A solution is to modify `with_numpy_cupy`: remove the output named arg and instead check if it is in the `**kwargs`. You would get a generic enough function to suit your  (and potentially other's) needs.

tests/operators/test_offres_exp_approx.py
```diff
@@ -0,0 +1,169 @@
...
+
+if AUTOGRAD_AVAILABLE:
+    params.extend([(2, "torch", "cpu"), (3, "torch", "cpu")])
+
+if AUTOGRAD_AVAILABLE and CUPY_AVAILABLE:
+    params.extend([(2, "torch", "gpu"), (3, "torch", "gpu")])
+
+
+@pytest.fixture(scope="module", params=params)
+def map_fixture(request):
```
paquiteau, line 96: A few suggestions:
- Just name it b0_map, the fixture suffix is redundant.
- Use `params_array_interface` decorator
- parametrize the fixture/TestCase directly (see the `operator` one in `test_batch.py` or `test_interface.py` for instance

tests/operators/test_offres_exp_approx.py
```diff
@@ -0,0 +1,169 @@
...
+    """Calculate approximate off-resonance modulation term."""
+    field_term = 0.0
+    for n in range(B.shape[0]):
+        tmp = B[n] * C[n][..., None]
+        tmp = tmp[None, ...].swapaxes(0, -1)[..., 0]
+        field_term += tmp
+    return field_term
+
+
+def _make_b0map(ndim, npix=64):
```
paquiteau, line 38: you could probably refactor the three function below into a Case (like the CaseTrajectory) in `tests/case_trajectories.py`

tests/operators/_test_offres_exp_approx.py
```diff
@@ -0,0 +1,236 @@
...
+def assert_allclose(actual, expected, atol, rtol):
+    """Backend agnostic assertion."""
+    xp = get_array_module(actual)
+    if xp.__name__ == "numpy":
+        npt.assert_allclose(actual, expected, atol, rtol)
+    if xp.__name__ == "cupy":
+        npt.assert_allclose(actual.get(), expected.get(), atol, rtol)
+    if xp.__name__ == "torch":
+        npt.assert_allclose(
+            actual.numpy(force=True), expected.numpy(force=True), atol, rtol
```
paquiteau, line 134: you can instead use the `from_interface` helper, directly in the test. otherwise, if the test is failing you don't get a nice traceback

paquiteau (changes requested):
Thanks for pushing it further ! It seems that you have duplicated the `test_offres_exp_approx.py` file. I have reviewed the non underscore function. @Daval-G will probably review next week , and we will be good to merge I guess :smile:

examples/example_offresonance.py
mcencini: With brainweb-dl mri_data != 0 is indeed not working - I am now using the more robust version

src/mrinufft/operators/off_resonance.py
mcencini: Ok, this is a bit more complicated than I expected. The decorator is mainly designed for class methods, and expect that "self" is passed. For standalone functions, "data" will be the second argument of the wrapper. I could:

1) move `get_interpolators_from_fieldmap` inside MRIFourierCorrected.
2) modify the decorator to inspect the signature and see if "self" or "cls" are detected as first argument, and branch between the class and standalone versions
3) add a "classmethod = False" argument to the decorator and switch between the two brances according to this flag. This would be more robust than 2) as both self/cls are conventions and might not be always respected - setting the default flag as False should make it backward compatible with the existing code.

1) is the cleanest, but maybe it would be nice to be able to decorate standalone functions? what do you think?

paquiteau: You raise a valid point. I would say 2) would be the long term solution (using `inspect.ismethod()` , and refactoring all the `with_` decorators we have build). This is not the purpose of this PR, so let's just focus on the B0 stuff for now, and keep your working code as is.

examples/example_offresonance.py
mcencini: Done, thanks!

tests/case_fieldmaps.py
```diff
@@ -0,0 +1,57 @@
...
+        """Create a real (B0 only) 2D field map."""
+        return make_b0map(2 * [N])
+
+    def case_real3D(self, N=64, b0range=(-300, 300)):
+        """Create a real (B0 only) 3D field map."""
+        return make_b0map(3 * [N])
+
+
+class CasesZmaps:
+    """Complex zmap field maps cases we want to test.
```
chaithyagr, line 22: Just a question, what do you mean by zmap here?

src/mrinufft/extras/data.py
chaithyagr, line 1: Maybe we call this `field_map.py`? @paquiteau

chaithyagr (approved):
Really minor things on  my side, otherwise we can merge soon! Thank you again for your enthusiasm, interest and codes!

tests/case_fieldmaps.py
mcencini: I was following the notation from this paper from [Fessler and Noll](https://deepblue.lib.umich.edu/bitstream/handle/2027.42/85903/Fessler50.pdf?sequence=1&isAllowed=y). Happy to change it to something else if you think this is confusing!

chaithyagr: No I think I am okay with it, just wanted to know the reasoning

chaithyagr:
Rebased codes to master, also building docs to ensure they are good. Finally I have added the single line change from #187 . To me we can merge when green (need the tests to run as gpuNUFFT was recently updated with critical memory leak fixes). Ill also wait on @paquiteau on any final remarks.

Great work @mcencini, thank you so much for it !

mcencini:
Thank you very much for the review! I should have fixed the failing finufft test (now I am skipping torch-gpu if cupy is not available), but I am not sure about what could be wrong in the gpuNUFFT one)

mcencini:
Apparently not :(  could it be related to gpunufft update?

src/mrinufft/operators/off_resonance.py
mcencini: Ok, I have done it in a different branch, I'll open a PR after this is merged

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    and MIRT (mri_exp_approx.m): https://web.eecs.umich.edu/~fessler/code/
+
+    Parameters
+    ----------
+    fieldmap : np.ndarray
+        Rate map defined as ``fieldmap = R2*_map + 1j * B0_map``.
+        ``*_map`` and ``t`` should have reciprocal units.
+        If ``zmap`` is real, assume ``zmap = B0_map``.
+        Also supports Cupy arrays and Torch tensors.
+        Expected shape is ``(nz, ny, nx)``.
```
Daval-G, line 40: Why are the dimensions expected to be in that order exactly ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+
+    From Sigpy: https://github.com/mikgroup/sigpy
+    and MIRT (mri_exp_approx.m): https://web.eecs.umich.edu/~fessler/code/
+
+    Parameters
+    ----------
+    fieldmap : np.ndarray
+        Rate map defined as ``fieldmap = R2*_map + 1j * B0_map``.
+        ``*_map`` and ``t`` should have reciprocal units.
+        If ``zmap`` is real, assume ``zmap = B0_map``.
```
Daval-G, line 38: `zmap` was not introduced in the documentation. I believe you refer to `fieldmap` from the same sentence.

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+if AUTOGRAD_AVAILABLE:
+    import torch
+
+
+def get_interpolators_from_fieldmap(
+    fieldmap, readout_time, n_time_segments=6, n_bins=(40, 10), mask=None
+):
+    r"""Create B and C matrices to approximate ``exp(-2j*pi*fieldmap*t)``.
+
+    Here, B has shape ``(n_time_segments, len(t))``
```
Daval-G, line 27: Could you explicit the relation between `t` and `readout_time` somewhere ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    and C has shape ``(n_time_segments, *fieldmap.shape)``.
+
+    From Sigpy: https://github.com/mikgroup/sigpy
+    and MIRT (mri_exp_approx.m): https://web.eecs.umich.edu/~fessler/code/
+
+    Parameters
+    ----------
+    fieldmap : np.ndarray
+        Rate map defined as ``fieldmap = R2*_map + 1j * B0_map``.
```
Daval-G, line 36: I wonder, isn't there a world where we want to correct only based on `R2*_map` ? The current parameters don't allow the user to do it simply as if no imaginary B0 is given, the R2* map will be assumed to be B0. So why not just require explicitly R2*_map and B0_map separately ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+def get_interpolators_from_fieldmap(
+    fieldmap, readout_time, n_time_segments=6, n_bins=(40, 10), mask=None
+):
+    r"""Create B and C matrices to approximate ``exp(-2j*pi*fieldmap*t)``.
+
+    Here, B has shape ``(n_time_segments, len(t))``
+    and C has shape ``(n_time_segments, *fieldmap.shape)``.
+
+    From Sigpy: https://github.com/mikgroup/sigpy
+    and MIRT (mri_exp_approx.m): https://web.eecs.umich.edu/~fessler/code/
```
Daval-G, line 31: As mentioned by @paquiteau, the approach based on SVD is optimal and therefore superior to the time segmented approach (at least with respect to this particular problem). Is there a reason to stick to the time segmented method ?

Reference to the SVD-based method: Fessler, Jeffrey A., Sangwoo Lee, Valur T. Olafsson, Hugo R. Shi, and Douglas C. Noll. "Toeplitz-based iterative image reconstruction for MRI with correction for magnetic field inhomogeneity." IEEE Transactions on Signal Processing 53, no. 9 (2005): 3393-3402.

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    import torch
+
+
+def get_interpolators_from_fieldmap(
+    fieldmap, readout_time, n_time_segments=6, n_bins=(40, 10), mask=None
+):
+    r"""Create B and C matrices to approximate ``exp(-2j*pi*fieldmap*t)``.
+
+    Here, B has shape ``(n_time_segments, len(t))``
+    and C has shape ``(n_time_segments, *fieldmap.shape)``.
```
Daval-G, line 28: @paquiteau @chaithyagr that shape makes the matrix quite memory intensive but it is absolutely necessary for R2* + B0 map correction at the same time. During my PhD I just avoided that because it would have been impossible to store on GPU. Should we propose a B0-only less intensive version based on the pysap-mri implementation ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+        Readout time in ``[s]`` of shape ``(nshots, npts)`` or ``(nshots * npts,)``.
+        Also supports Cupy arrays and Torch tensors.
+    n_time_segments : int, optional
+        Number of time segments. The default is ``6``.
+    n_bins : int | Sequence[int] optional
+        Number of histogram bins to use for ``(B0, T2*)``. The default is ``(40, 10)``
+        If it is a scalar, assume ``n_bins = (n_bins, 10)``.
+        For real fieldmap (B0 only), ``n_bins[1]`` is ignored.
+    mask : np.ndarray, optional
+        Boolean mask to avoid histogram of background values.
```
Daval-G, line 51: Please make the description more user-oriented, like is it a mask of the background or the ROI ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    import cupy as cp
+
+if AUTOGRAD_AVAILABLE:
+    import torch
+
+
+def get_interpolators_from_fieldmap(
+    fieldmap, readout_time, n_time_segments=6, n_bins=(40, 10), mask=None
+):
+    r"""Create B and C matrices to approximate ``exp(-2j*pi*fieldmap*t)``.
```
Daval-G, line 25: Please specify that B and C are the low rank approximation of the matrix defined by this formula, the relation is not very clear even if we expect users to be a bit familiar

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    C : np.ndarray
+        Off-resonance phase map at each time segment center of shape
+        ``(n_time_segments, *fieldmap.shape)``.
+        Array module is the same as input fieldmap.
+    """
+    # default
+    if isinstance(n_bins, (list, tuple)) is False:
+        n_bins = (n_bins, 10)
+
+    # transform to list
```
Daval-G, line 69: No need to comment such operations. It is generally considered a bad habit to comment too much, as readers will just stop reading the comments if they are too obvious

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    xp = get_array_module(fieldmap)
+
+    # enforce complex field
+    fieldmap = xp.asarray(fieldmap, dtype=xp.complex64)
+
+    # cast arrays to fieldmap backend
+    if xp.__name__ == "torch":
+        is_torch = True
+    else:
+        is_torch = False
```
Daval-G, line 82:
```suggestion
    is_torch = xp.__name__ == "torch"
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    if mask is None:
+        mask = xp.ones_like(fieldmap, dtype=bool)
+    else:
+        mask = xp.asarray(mask, dtype=bool)
+
+    # get field map
+    if xp.isreal(fieldmap).all().item():
+        r2star = None
+        b0 = fieldmap
+        fieldmap = 0.0 + 1j * b0
```
Daval-G, line 106:
```suggestion
        fieldmap = 1j * b0
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    else:
+        r2star = fieldmap.real
+        b0 = fieldmap.imag
```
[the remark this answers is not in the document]
Daval-G: and r2star is only used as a boolean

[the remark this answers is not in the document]
Daval-G: I don't get the point of that block

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # get bin centers
+        zc = [e[1:] - (e[1] - e[0]) / 2 for e in ze]
+
+        # complexify
+        zk = _outer_sum(1j * zc[0], zc[1])  # [K1 K2]
+        zk = zk.T
+        hk = hk.T
+    else:
+        z = fieldmap[mask].ravel()
```
Daval-G, line 129: This line can be factorized out of the if statement

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+        hk, ze = xp.histogram(z.imag, bins=n_bins[0])
+
+        # get bin centers
+        zc = ze[1:] - (ze[1] - ze[0]) / 2
+
+        # complexify
+        zk = 0 + 1j * zc  # [K 1]
```
Daval-G, line 136:
```suggestion
        zk = 1j * zc  # [K 1]
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    B = B.astype(xp.complex64)
+
+    # get spatial coeffs
+    C = xp.exp(-tl * fieldmap[..., None])
+    C = C[None, ...].swapaxes(0, -1)[
+        ..., 0
+    ]  # (..., n_time_segments) -> (n_time_segments, ...)
+
+    # clean-up of spatial coeffs
+    C = xp.nan_to_num(C, nan=0.0, posinf=0.0, neginf=0.0)
```
Daval-G, line 163: Should `C` be also casted to np.complex64, as for `B` ?

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    fourier_op: object of class FourierBase
+        the fourier operator to wrap
+    fieldmap : np.ndarray, optional
+        Rate map defined as ``fieldmap = R2*_map + 1j * B0_map``.
+        ``*_map`` and ``t`` should have reciprocal units.
+        If ``zmap`` is real, assume ``zmap = B0_map``.
+        Also supports Cupy arrays and Torch tensors.
+        Expected shape is ``(nz, ny, nx)``.
+    readout_time : np.ndarray or GPUarray, optional
+        Readout time in ``[s]`` of shape ``(nshots, npts)`` or ``(nshots * npts,)``.
```
Daval-G, line 227:
```suggestion
        Readout time in ``[s]`` of shape ``(n_shots, n_pts)`` or ``(n_shots * n_pts,)``.
```

Please keep a consistent naming convention  as in `n_time_segments` or `n_bins`

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+class MRIFourierCorrected(FourierOperatorBase):
+    """Fourier Operator with B0 Inhomogeneities compensation.
+
+    This is a wrapper around the Fourier Operator to compensate for the
+    B0 inhomogeneities  in the  k-space.
+
+    Parameters
+    ----------
+    fourier_op: object of class FourierBase
+        the fourier operator to wrap
```
Daval-G, line 219:
```suggestion
        the Fourier operator to wrap
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+# */
+
+
+class MRIFourierCorrected(FourierOperatorBase):
+    """Fourier Operator with B0 Inhomogeneities compensation.
+
+    This is a wrapper around the Fourier Operator to compensate for the
+    B0 inhomogeneities  in the  k-space.
```
Daval-G, line 214:
```suggestion
    This is a wrapper around the Fourier Operator to compensate for the
    B0 inhomogeneities in  the k-space.
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Parameters
+    ----------
+    fourier_op: object of class FourierBase
+        the fourier operator to wrap
+    fieldmap : np.ndarray, optional
+        Rate map defined as ``fieldmap = R2*_map + 1j * B0_map``.
+        ``*_map`` and ``t`` should have reciprocal units.
+        If ``zmap`` is real, assume ``zmap = B0_map``.
```
Daval-G, line 223: Same as above

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+    n_time_segments : int, optional
+        Number of time segments. The default is ``6``.
+    n_bins : int | Sequence[int] optional
+        Number of histogram bins to use for ``(B0, T2*)``. The default is ``(40, 10)``
+        If it is a scalar, assume ``n_bins = (n_bins, 10)``.
+        For real fieldmap (B0 only), ``n_bins[1]`` is ignored.
+    mask : np.ndarray or GPUarray, optional
+        Boolean mask to avoid histogram of background values.
+        The default is ``None`` (use the whole map).
+        Also supports Cupy arrays and Torch tensors.
```
Daval-G, line 238: Same as above

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+        else:
+            raise ValueError("Unsupported backend.")
+        self._fourier_op = fourier_op
+
+        self.n_coils = fourier_op.n_coils
+        self.shape = fourier_op.shape
+        self.smaps = fourier_op.smaps
+        self.autograd_available = fourier_op.autograd_available
+
+        if B is not None and C is not None:
```
Daval-G, line 278:
```suggestion
        if B and C:
```

Daval-G (changes requested):
Thank you for your contributions. I have mostly minor comments on the code here and there. Sorry to review so late, I was on vacation and I don't work on MRI-nufft on my working hours.

Daval-G:
> Ahh dang. That's on me, let me fix that. Don't worry about it, I'll merge once everything is green

Haha no

chaithyagr:
> > Ahh dang. That's on me, let me fix that. Don't worry about it, I'll merge once everything is green
> 
> Haha no

Thats good for me, while @mcencini handles your review, I get some time to fix gpuNUFFT xD

src/mrinufft/operators/off_resonance.py
mcencini: Thanks! I initially accepted, but then reverted because, since B and C are arrays / tensors when provided, this line would return:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

chaithyagr:
I am just double checking the gpuNUFFT tests. Locally they all seem to pass fine..

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,330 @@
...
+        xp = np
+        fieldmap = _to_numpy(fieldmap)
+        readout_time = _to_numpy(readout_time)
+        mask = _to_numpy(mask)
+
+    readout_time = xp.asarray(readout_time, dtype=xp.float32).ravel()
+    if mask is None:
+        mask = xp.ones_like(fieldmap, dtype=bool)
+    else:
+        mask = xp.asarray(mask, dtype=bool)
```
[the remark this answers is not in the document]
mcencini: Same as below, this would raise a ValueError [...] use mask.any() or mask.all()

src/mrinufft/operators/off_resonance.py
mcencini: You are completely right, the only important thing is that the order is consistent between field map and image

src/mrinufft/operators/off_resonance.py
mcencini: You are completely right, sorry for that

src/mrinufft/operators/off_resonance.py
mcencini: I agree!

src/mrinufft/operators/off_resonance.py
mcencini: Done!

src/mrinufft/operators/off_resonance.py
mcencini: Done

src/mrinufft/operators/off_resonance.py
mcencini: Done, thanks!

src/mrinufft/operators/off_resonance.py
mcencini: You are right, I have now refactored the function which now accepts separate B0 and R2* (also easier for the user, which does not have to manually create the complex field map.

src/mrinufft/operators/off_resonance.py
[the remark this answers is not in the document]
mcencini: You are right, I removed it during the fieldmap -> (b0, r2*) refactor

src/mrinufft/operators/off_resonance.py
mcencini: Good idea!

src/mrinufft/operators/off_resonance.py
mcencini: Yes, that was a mistake. Superseded by the refactor

src/mrinufft/operators/off_resonance.py
mcencini: Removed during the refactor

src/mrinufft/operators/off_resonance.py
mcencini: Also superseded during the refactor

src/mrinufft/operators/off_resonance.py
mcencini: I hopefully improved the description

src/mrinufft/operators/off_resonance.py
mcencini: Ok, I had a look to pysap-mri offresonance wrapper. If I understand it correctly, during the operator application it loops over the field coefficients (like my implementation) and it recreate the l-th spatial interpolator for each iteration in the loop:

```
 for l in range(self.num_interpolators):
            y += self.B[..., l] * self.fourier_op.op(
                self.C[l, self.indices] * x,
                *args
            )
```

Memory-wise, this is equal to compute "online" the l-th spatial interpolator as `xp.exp(-1j * tl[l] * field_map)`. I have done [some tests](https://gist.github.com/mcencini/b872bc74f57c52762953f533a57ed9f5) to compare the runtimes of indexing (`self.C[l, self.indices]`) and exp operation for different backends, and these are my results:

```
numpy indexing:
55.3 ms ± 110 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
numpy exp:
330 ms ± 387 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
cupy indexing:
475 μs ± 15.6 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
cupy exp:
58.7 μs ± 17.2 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
torch cpu indexing:
13.9 ms ± 1.68 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
torch cpu exp:
54.9 ms ± 318 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
torch cuda indexing:
478 μs ± 5.89 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
torch cuda exp:
31.2 μs ± 15.7 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Apparently, for CPU (mine is a Intel(R) Xeon(R) Gold 6342 CPU @ 2.80GHz with 24 cores) indexing is faster than exponential, while for GPU (NVIDIA A40) I observe the opposite (exponential is faster). Since storing the whole `(n_time_segments, *shape)` matrix should not be a problem on CPU, I think we could switch to online computation of `C[idx]` if `is_device_array(b0_map)` is True.

mcencini:
I wanted to test locally as well but pip tries to build gpuNUFFT and cmake does not find CUDA, so it will take a while unfortunately

chaithyagr:
Don't worry about it, worse case we just remove the breaking change for
now... Once I get green light if gpunufft issue isn't fixed, I'll just
revert it

mcencini:
Cufinufft passes the tests, locally

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 
-gpunufft = ["gpuNUFFT>=0.9.0", "cupy-cuda12x"]
+gpunufft = ["gpuNUFFT>=0.10.0", "cupy-cuda12x"]
```
chaithyagr, line 14:
```suggestion
gpunufft = ["gpuNUFFT>=0.9.0", "cupy-cuda12x"]
```

chaithyagr:
Reverted back gpuNUFFT. Ill wait for green light on tests and from finally from @Daval-G and we are finally done!

chaithyagr:
@mcencini I see some failures on some tests you proposed: 

tests/operators/test_offres_exp_approx.py .......F.......F               [ 46%]

Can you please take a look? @paquiteau can you have a look at the stack one in your free time? Its weird new issue, did something change at master recently?

mcencini:
Ouch, I did not see that... I'll try to create the same enivironment of the runner to see if I can reproduce

mcencini:
I was not able to reproduce in a fresh Python 3.10 environment... it is a bit difficult to understand the error as the segfault in test_stacked prevents (I think) a meaningful report from pytest. Would it make sense (also to save resources) to add the -x flag in the pytest command on the runner so that the test stops at the first failure?

paquiteau (approved):
Tests are flakky right now, I would put that on our CI setup (and the local runner we use at the lab). This should be addressed separately from this PR.

src/mrinufft/operators/off_resonance.py
Daval-G: You did great ! Also you made me realize I wrongly believed it was not possible because of the 2d-histogram to do it in the case of B0+R2* correction. I agree with your suggestions.

On another note, it reminded me of some tests I made with binning precision. The default parameter is (40, 10), how was it chosen exactly ? We could check directly on a typical 3T or 7T field map + R2* map the matrix factorization error depending on the number of bins to suggest the default arguments if not already done, what do you think ?

src/mrinufft/operators/off_resonance.py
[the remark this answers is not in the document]
Daval-G: You are right, my bad !

src/mrinufft/operators/off_resonance.py
mcencini: Thank you very much! :D I used the same default as MIRT, but I don't know how this was estimated. Unfortunately I do not dispose of typical 3T / 7T complex field maps right now - maybe we could use the same as your pysap-mri implementation?
```
if (n_bins == "auto"):
            n_bins = 2 * int(np.around(range_w[1] - range_w[0]))
```

Daval-G: We can just create an issue to make that study later. I would advise against using the method in my pysap-mri implementation, as it was a rule of thumb that just account for the off-resonance range, while it should also be dependent on the readout length. We can craft a better rule based on a later study

mcencini: Sounds good to me!

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,428 @@
...
+    B : np.ndarray
+        Temporal interpolator of shape ``(n_time_segments, len(t))``.
+        Array module is the same as input field_map.
+    tl : np.ndarray
+        Time segment centers of shape ``(n_time_segments,)``.
+        Array module is the same as input field_map.
+
+    """
+    # default
+    if isinstance(n_bins, (list, tuple)) is False:
```
Daval-G, line 82:
```suggestion
    if not isinstance(n_bins, (list, tuple)):
```

src/mrinufft/operators/off_resonance.py
```diff
@@ -0,0 +1,428 @@
...
+
+    if xp.__name__ == "numpy":
+        return torch.from_numpy(input)
+    elif xp.__name__ == "cupy":
+        return torch.from_dlpack(input)
+    else:
+        return input
+
+
+# */
```
Daval-G, line 206: What is that ?

Daval-G (approved):
Just two small comments, besides that it seems good to merge on my side

src/mrinufft/operators/off_resonance.py
mcencini: I just wanted to mark the block I will remove after the with_* decorators (e.g., with_torch) will be generalized (I have done some work on this and will open another PR after this one is ok)

paquiteau:
So let's go guys ? we merge this !

chaithyagr:
Hopefully now, we get all green and soon we merge! Thank you again @mcencini !

PR #185, opened 2024-08-29
add Conjugate Gradient with Momentum  

Lenoush:
In base.py , add a first draft of cg function with 
`
cg(self, kspace_data, x_init=None, num_iterations=10, tol=1e-4)
`
where x_init = initial image , tol = tolerance   

Classique example of gradient descent  : 
```
def cg(fourier_op, kspace_data, num_iterations):
    L = fourier_op.get_lipschitz_cst()
    image = np.zeros(fourier_op.shape)
    for i in range(num_iterations):
         grad = fourier_op.data_consistency(image, kspace_data)
         image = image - grad * L
    return image
```  

Help of :
- https://sigpy.readthedocs.io/en/latest/_modules/sigpy/alg.html#ConjugateGradient
- https://aquaulb.github.io/book_solving_pde_mooc/solving_pde_mooc/notebooks/05_IterativeMethods/05_02_Conjugate_Gradient.html#


paquiteau:
I think this would be more suited in `mrinufft.extras`

chaithyagr:
@Lenoush I think I agree with @paquiteau, can we move this to extras module. Also it would help to have test and an example?

examples/example_cg.py
```diff
@@ -0,0 +1,35 @@
...
+kspace_data = nufft.op(image)  # get the k-space data
+reconstructed_image = cg(nufft, kspace_data)  # reconstruct the image
+
+
+# Display the results
+plt.figure(figsize=(10, 5))
+plt.subplot(1, 2, 1)
+plt.title("Original Image")
+plt.imshow(abs(image), cmap="gray")
+
```
paquiteau, line 31: Also add the single step adjoint image, to show that the conjugate gradient does better (hopefully)

paquiteau (changes requested):
This looks very good ! 
The example deserves more explanation (add a ref to the CG algorithm, Wikipedia is okay)

examples/example_cg.py
Lenoush: not sure the conjugate gradient is better

paquiteau (approved):
LGTM

PR #194, opened 2024-09-17
Generalize with_* decorators

mcencini:
This PR generalizes the `with_*` decorators to standalone functions, as discussed in PR #180.


src/mrinufft/operators/base.py
```diff
@@ -115,136 +117,284 @@ class or instance of class if args or kwargs are given.
...
+
     return operator
 
 
 def with_numpy(fun):
     """Ensure the function works internally with numpy array."""
+    if _ismethod(fun):
+        data_arg_idx = 1
+    else:
+        data_arg_idx = 0
```
paquiteau, line 129:
```suggestion
    data_arg_idx = _ismethod(fun)
```
boolean, integers, its the same for python

src/mrinufft/operators/base.py
```diff
@@ -115,136 +117,284 @@ class or instance of class if args or kwargs are given.
...
+        return get_array_module(input)
+    except Exception:
+        return np
+
+
+def _get_device(input):
+    if is_cuda_array(input) or is_cuda_tensor(input):
+        return input.device
+    else:
+        return None
```
paquiteau, line 248: numpy 2.0 has added support for array.device this is the   [array_api](https://data-apis.org/array-api/latest/API_specification/array_object.html#attributes) standard 
I don't think you need this anymore :)

src/mrinufft/operators/base.py
```diff
@@ -115,136 +117,284 @@ class or instance of class if args or kwargs are given.
...
 
     return wrapper
 
 
+def _ismethod(fun):  # ismethod works on instance methods, not classes (always False)
+    first_arg = list(inspect.signature(fun).parameters)[0]
+    if first_arg in ["self", "cls"]:
+        return True
+    else:
+        return False
```
paquiteau, line 234:
```suggestion
    return first_arg in ["self", "cls"]:
```

src/mrinufft/operators/base.py
```diff
@@ -115,136 +117,284 @@ class or instance of class if args or kwargs are given.
...
+
+
+def _get_array_module(input):  # handle native Python case
+    try:
+        return get_array_module(input)
+    except Exception:
+        return np
```
paquiteau, line 241: I think we could live happily with `get_array_module` returning numpy in the default case as well (and forget about the exception altogether)

paquiteau (changes requested):
Thanks for the PR :)  I have found some places to improve but this is a great start ! More general comments:
- I think all these wrapper / array conversion could be refactor in its own module (`mrinufft/_array_compat.py` ? ) 
- Add some docstrings they are still useful, even for private functions :) 
- Some tests on top of that would be perfect !

src/mrinufft/operators/base.py
mcencini: Here I want to select the index of the old "data" argument assuming it is the first (for standalone functions and staticmethods) or second (class or instance methods) positional argument. Using booleans, `args[False]` would return `[]`, while `args[True]` is the same as `args[None,...]` (apparently). If we want to avoid the `if - else` statement, we should do something like `data_arg_idx = int(not(_ismethod(fun))`, but I am not sure it is that readable...

src/mrinufft/operators/base.py
mcencini: Ahhh, I was not aware of that, thanks! I noticed that mri-nufft requires "numpy<2.0" right now, should we keep compatibility with numpy 1.x for now? I think it makes sense, probably there are many packages still transitioning.

P.S: I also noticed that the optional pynfft dependency requires numpy>=2.0.0, this seems conflicting with the mandatory dependency

src/mrinufft/operators/base.py
paquiteau: Ah good catch ! I am okay with `data_arg_idx = int(not(_ismethod(fun))` :)

src/mrinufft/operators/base.py
paquiteau: In this case, you can get away with a try/except and return "cpu" if there is an exception

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 
-gpunufft = ["gpuNUFFT>=0.9.0", "cupy-cuda12x"]
```
chaithyagr, line 14: Maybe we dont want this change. It got passed it from one of my commits, but which was later removed.

chaithyagr (changes requested):
Thank you for this!

src/mrinufft/operators/base.py
mcencini: My bad, args is obviously a list and not a numpy array, so indexing with booleans would actually work as intended (also, I have no idea why I thought I needed a "not", sorry)

chaithyagr (approved):
LGTM, Ill let @paquiteau to finalize this PR... Thank you @mcencini .

src/mrinufft/_array_compat.py
```diff
@@ -0,0 +1,379 @@
...
+        return input.device
+    except Exception:
+        return "cpu"
+
+
+def _get_args(func, args, kwargs):
+    """Convert input args/kwargs mix to a list of positional arguments.
+
+    This automatically fills missing kwargs with default values.
+    """
```
paquiteau, line 158: The fact that it only returns positional arguments is tricky. if a wrapped function as **kwargs in its signature, the named parameters aspect is lost (which can be detrimental if their are handled further down in the wrapped function). I would advise to return args and kwargs separately. 

This would implies that the `_to_<array_module>` functions should process both args and kwargs

tests/test_array_compat.py
```diff
@@ -0,0 +1,140 @@
...
+        assert isinstance(
+            output, expected_type
+        ), f"Expected {expected_type} but got {type(output)}"
+        assert (
+            _get_device(output) == expected_device
+        ), f"Expected {expected_device} but got {_get_device(output)}"
+
+
+@_param_array_interface
+def test_internal_conversions(decorator_factory, array_interface):
```
paquiteau, line 77: I like this,  but I think it could be refactor/parametrize  properly. An attempt:

```python
@paramtrize(with_decorator=[with_numpy,  ... ]) # probably needs some skips as well 
@parametrize("input_array_interface, array_fun") = [("numpy", "numpy.array"), ("cupy","cupy.array"), ...])
def test_internal_conversion(with_decorator, input_array_interface, array_fun):

    # Create input array
    array = to_interface(np.asarray([1.0, 2.0, 3.0]), array_interface)
    
    # Execute function and monitor internal conversion
    with patch(array_fun, wraps=eval(array_fun)) as mock_fun:
        _  = with_decorator(dummy_fun)(array, array, array, 42, "dummy")
        assert mock_fun.call_count == 3

```

An other solution would be write a test for each specific case (at least it would be more explicit which case failed in the testing)

paquiteau (changes requested):
Some thoughts about the testing, this is not trivial

src/mrinufft/_array_compat.py
mcencini: At the beginning, I tried to handle both args and kwargs. The main problem I encountered was that if we pass only keyworded arguments, then things like `xp = get_array_module(args[data_arg_idx])` fail because there are no positional arguments left. The issue you are raising is indeed very valid - any suggestion on how to handle it? What about iterating over args and kwargs and picking the first arraylike argument as leading?

src/mrinufft/_array_compat.py
```diff
@@ -0,0 +1,463 @@
...
+            elif "tensorflow" in xp.__name__:
+                if "CPU" in _arg.device:
+                    with cp.cuda.Device(device):
+                        _arg = cp.asarray(_arg.numpy())
+                else:
+                    _arg = cp.from_dlpack(tf.experimental.dlpack.to_dlpack(_arg))
+        if isinstance(_arg, (tuple, list)):
+            _arg, _ = _to_cupy(*_arg, device=device)
+
+        kwargs[key] = _arg
```
paquiteau, line 270: This is redundant with the above, lets refactor:
```suggestion
    # convert keyworded 
    process_kwargs_vals = _to_cupy(*kwargs.values())

    kwargs = {k: v for k,v in zip(kwargs.keys(), process_kwargs_vals}
```

src/mrinufft/_array_compat.py
```diff
@@ -0,0 +1,463 @@
...
+                    _arg = torch.as_tensor(cp.asnumpy(_arg))
+            elif "tensorflow" in xp.__name__:
+                if "CPU" in _arg.device:
+                    _arg = torch.as_tensor(_arg.numpy(), device=device)
+                else:
+                    _arg = torch.from_dlpack(tf.experimental.dlpack.to_dlpack(_arg))
+        if isinstance(_arg, (tuple, list)):
+            _arg, _ = _to_torch(*_arg, device=device)
+
+        kwargs[key] = _arg
```
paquiteau, line 346: Apply same refactor trick

src/mrinufft/_array_compat.py
```diff
@@ -0,0 +1,463 @@
...
+                elif TF_CUDA_AVAILABLE:
+                    _arg = tf.experimental.dlpack.from_dlpack(
+                        torch.utils.dlpack.to_dlpack(_arg)
+                    )
+                else:
+                    _arg = tf.convert_to_tensor(_arg.numpy(force=True))
+        if isinstance(_arg, (tuple, list)):
+            _arg, _ = _to_tensorflow(*_arg)
+
+        kwargs[key] = _arg
```
paquiteau, line 420: apply same refactor trick

src/mrinufft/_array_compat.py
```diff
@@ -0,0 +1,463 @@
...
+            if xp.__name__ == "torch":
+                _arg = _arg.numpy(force=True)
+            elif xp.__name__ == "cupy":
+                _arg = cp.asnumpy(_arg)
+            elif "tensorflow" in xp.__name__:
+                _arg = _arg.numpy()
+        if isinstance(_arg, (tuple, list)):
+            _arg, _ = _to_numpy(*_arg)
+        kwargs[key] = _arg
+
```
paquiteau, line 204: apply the same refactor trick

tests/test_array_compat.py
```diff
@@ -0,0 +1,178 @@
...
+    @decorator
+    def test_func(*args):
+        return dummy_func(*args)
+
+    return test_func, decorator
+
+
+@_param_array_interface
+def test_decorators_outcome(decorator_factory, array_interface):
+    decorated_function, _ = decorator_factory
```
paquiteau, line 75: if you don't need the second argument, don't return it

paquiteau (changes requested):
I found a nice trick to avoid repeating codes ;) 

Otherwise this is good stuff !

paquiteau (approved):
LGTM! Great job @mcencini

Issue #118, opened 2024-05-26
Trajectory gallery: let's discuss


paquiteau:
From the discussions in #105, #106, #113, and more, It seems that a better display of the trajectories in MRI-NUFFT documentation (building on top of the amazing work of @Daval-G ). As more and more trajectories are added, the question of their discoverability is also at hand. 

Technically speaking, the Sphinx Gallery seems to be a good base to work on; an inspiring example can be from `matplotlib` itself (https://matplotlib.org/stable/gallery/index.html)! Using sections and tagging may be a good idea as well. 

Writing a script for each trajectory may be too repetitive, so maybe some templating system can be used (one solution may be to describe the trajectory in `.json`/`.yaml` and generate a `.py` file from it with [jinja](https://jinja.palletsprojects.com/en/3.0.x/templates/) ). Also, some extra CI configuration could be helpful.

However, it is necessary first to discuss exactly what we want to display for each trajectory.

Daval-G:
Overall I support the idea of generating automatically part of what we are doing by hand. The current examples are huge lists of unrelated argument descriptions, which is redundant with the documentation that just lacks the visualization part. I would see more structured examples, showing a more practical usage or discussing relation between trajectories. WDYT ?

Issue #182, opened 2024-08-24
Use stdlib tempfile module for examples


paquiteau:
Python built-in module [tempfile](https://docs.python.org/3/library/tempfile.html) is the canonical way of making tempfile. It would be nicer to use it in the examples that generate images (`example_nD_gif.py`, `example_learn_trajectory.py`, etc) 

Using a [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) is probably the way to go

PR #196, opened 2024-09-20
Clean CPU examples

Daval-G:
This PR cleans the CPU examples in several ways:
- Change brain orientations to the medical standard
- Reduce blank spaces around images
- Remove some overlaps in figures
- Split script-specific code from the utils and data preparation
- Homogeneize parts and subparts names & structures
- Homogeneize one-line descriptions
- Homogeneize capitalizations, punctuations
- Apply isort to clean imports
- Fix a bunch of typos
- Fix paragraph irregularities
- Fix a few cross references
- Break lines sometimes
- Remove a few dead parts
- Remove many useless outputs
- Remove references to MFI in off-resonance operator (we don't do MFI but the opposite)

I will try to do the same for GPU examples later. Note that trajectory examples are barely touched, not because they are perfect but the opposite as I plan to break them into smaller examples as discussed in #118.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -78,6 +78,7 @@ sphinx_gallery_conf = {
     "reference_url": {"mrinufft": None},
     "examples_dirs": ["../examples/"],
     "gallery_dirs": ["generated/autoexamples"],
+    "within_subsection_order": "ExampleTitleSortKey",
     "filename_pattern": "/example_",
     "ignore_pattern": r"(__init__|conftest|utils).py",
     "nested_sections": True,
--- a/examples/GPU/example_cg.py
+++ b/examples/GPU/example_cg.py
@@ -1,24 +1,27 @@
-# %%
 """
-Example of using the Conjugate Gradient method.
+======================================
+Reconstruction with conjugate gradient
+======================================
 
-This script demonstrates the use of the Conjugate Gradient (CG) method 
-for solving systems of linear equations of the form Ax = b, where A is a symmetric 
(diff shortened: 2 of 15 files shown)


paquiteau (approved):
Very nice improvements, thanks! 

If you wish, you can also tackle #182 in this as well.

paquiteau:
> By the way, how can we change the examples order in the gallery ? It doesn't make sense

RTFM :upside_down_face:  https://sphinx-gallery.github.io/stable/configuration.html#within-gallery-order

there are multiple possibilities; I guess `FileNameSortKey` and a proper naming scheme (`example_01_...`) could do the trick.

Daval-G:
> Very nice improvements, thanks!
> 
> If you wish, you can also tackle #182 in this as well.

Done

Daval-G:
> > By the way, how can we change the examples order in the gallery ? It doesn't make sense
> 
> RTFM 🙃 https://sphinx-gallery.github.io/stable/configuration.html#within-gallery-order
> 
> there are multiple possibilities; I guess `FileNameSortKey` and a proper naming scheme (`example_01_...`) could do the trick.

For now let's sort by example name rather than filename, until we decide of a specific order

Daval-G:
@paquiteau I cleaned the new CG example who seem to be problematic even on the main branch, but even after fix it doesn't work as intended on my side (image not showing because the algorithm doesn't converge). I moved it to GPU examples because it uses gpuNUFFT even though I doubt it is a requirement, but since it didn't work on my side I didn't want to change too many things

paquiteau (approved):
LGTM ! Thanks again :)

PR #195, opened 2024-09-20
Add support for density compensation estimation with cufinufft

chaithyagr:
This is crucial support for many use cases. I have some temporary things added to ensure tests still run good.

PR #206, opened 2024-11-01
Densities, random walks & travelling salesman

Daval-G:
This PR adds densities, sampling, and trajectories based on that (random walks and travelling salesman). In particular, it reproduces 3 papers from Nicolas Chauffert (4th coming "soon"). The features are represented in the images below, but in details:
- `maths.py` is split into a submodule as the file became too big
- functions were added to generate density arrays in 2D and 3D: cutoff-decay, polynomial, energy-based, Chauffert & fast Chauffert. The last one corresponds to an adaptation of the matlab code, but it uses only 1D vectors to compute the Fourier basis which appear to yield incomplete densities.
- `inits` submodule to store the trajectory initializations, because `trajectory2D.py` and `trajectory3D.py` are also starting to be too big
- Random walk trajectories in 2D and 3D based on any density
- Travelling Salesman trajectories in 2D and 3D with options to cluster shots before computing the TSP solution for time efficiency.

TODO:
- [x] Add docstrings
- [x] Write an example
- [x] Add references
- [x] Add `pywavelets` dependency

Density features:
density_update

Trajectory features:
chauffert_update


Daval-G:
@paquiteau I need to add a dependency to `pywavelets`, is it OK ? Should I prefer something else ? And should I add it as an optional dependency or not ?

paquiteau:
Whoa there is a lot to process, but this look awesome. @philouc or @chaithyagr you might want to take a look ;) 

For pywavelets, you can add it as an optional depedency (and give proper error message when not found) .

philouc:
@Daval-G did you make sure to define the optimal target sampling density (TSD) for TSP solutions as a function of the dimensions of the ambient space? In 2D (d=2), if \pi is supposed to be the optimal TSD, the TSD for TSP should be parameterized as \pi^2 as the TSP itself works as \pi^{d-1/d} [see section 4.3, p. 1976 in Chauffert et al, SIAM IS 2014] whereas in 3D the target TSD should be initialized as \pi^{3/2}.

Daval-G:
@philouc Good catch, I overlooked that point.

I followed the experiment proposed in the paper and presented in Fig 4.1, a.k.a I made 10 000 trajectories/sampling (Nc=10, Ns=100) in different situations and then made a 2D histogram to find if the target density was actually respected or not. I based that experiment on a dummy TSD as according to the paper it is not specific to Nicolas's optimal TSD.

In the figure below, (a) is the TSD and (b-f) are normalized histogram corresponding to the empirical distributions:
density_study

Note that:
- "Pseudo-random walk" (c) is just a random walk where the probability is adjusted dynamically to reduce the probability to go back to already covered areas. It is an option provided to the users for the random walk.
- "Random TSP x4" (f) means that the trajectory consisted of a Travelling Salesman (TS) based on random sampling where only 1/4 of points were drawn, and 3/4 of points are simply interpolated along the TS path/solution. I did that because otherwise it would be strictly equivalent to just random sampling.
- "Lloyd sampling" was actually sampled only 1000 times because it takes a lot more time than the other ones.

Overall it seems that the TSD is already respected without additional correction (except for (e) but it was not expected anyway). I am not sure why it is the case here and not in the paper, I am still looking into it.

src/mrinufft/trajectories/densities.py
paquiteau, line 1: It would be nice to have docstrings, for the modules and all the functions therein (with references to the papers)

src/mrinufft/trajectories/inits/random_walk.py
paquiteau, line 1: docstrings for the public functions :)

src/mrinufft/trajectories/maths/constants.py
paquiteau, line 1: I think you can just declare the constant in the `__init__.py` (they are not used in the child modules)

src/mrinufft/trajectories/maths/tsp_solver.py
```diff
@@ -0,0 +1,33 @@
+"""Solver for the Travelling Salesman Problem."""
+
+import numpy as np
+
+
+def solve_tsp_with_2opt(locations, improvement_threshold=1e-8):
```
paquiteau, line 6: A nice docstring here as well

paquiteau (changes requested):
A first pass, the code is looking good to me, I let @chaithyagr or @philouc comment on the maths. 

I guess you already plan to add more documentation and examples?

Also, some of the internal maths functions can be quite complex, so if you have some debugging setup that you wish to promote to unit-tests, feel free to do so as well.

Daval-G:
@paquiteau Yes I made a todo list in the original PR post about docstrings, examples & co. I converted it to draft

philouc:
You're right @Daval-G. Illustration on a dummy TSD is fine. I was just referring to the math relation allowing us to compute the sampling density on the TSP (points are joined by straight lines) vs the original TSD which could be sampled independently. Based on this comment I'm not sure whether TSPx4 shown in panel (f) fulfills the equation because if you start from the TSD in (a) you should end up with the sqrt of (a) in (f). Fig. 6 in Nicolas' paper illustrates pretty well that phenomemon when contrasting panes (d) and (g).

src/mrinufft/trajectories/densities.py
chaithyagr, line 1: Maybe add in mrinufft/densities ? @paquiteau your opinion?

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
+"""Sampling densities and methods."""
+
+import numpy as np
+import numpy.fft as nf
+import numpy.linalg as nl
+import numpy.random as nr
```
chaithyagr, line 6: I think this sacrifices a bit on readability. I would rather prefer np.random.<> np.linalg.<> etc..

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
...
+import pywt as pw
+from sklearn.cluster import BisectingKMeans, KMeans
+from tqdm.auto import tqdm
+
+from .utils import KMAX
+
+
+def sample_from_density(nb_samples, density, method="random"):
+    rng = nr.default_rng()
```
chaithyagr, line 15: Maybe it is worth mentioning on the docstring that it will still be samples on a grid.

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
...
+            sample_weight=density.flatten(),
+        )
+        locations = kmeans.cluster_centers_ - np.array(density.shape) / 2
+        locations = KMAX * locations / np.max(np.abs(locations))
+    else:
+        raise ValueError(f"Unknown sampling method {method}.")
+    return locations
+
+
+def create_cutoff_decay_density(shape, cutoff, decay, resolution=None):
```
chaithyagr, line 51: How about a more generic density defined in Eq. 10 in https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.29702 ? 
Its a very short update.
Also, having support for elliptical sampling enabled or not can be helpful (we soon plan to disable it :P )

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
...
+
+    density = np.zeros(shape)
+    unit_vector = np.zeros(shape)
+    for ids in tqdm(indices, disable=not verbose):
+        ids = tuple(ids)
+        unit_vector[ids] = 1
+        fourier_vector = nf.ifftn(unit_vector)
+        coeffs = pw.wavedecn(
+            fourier_vector, wavelet=wavelet_basis, level=nb_wavelet_scales
+        )
```
chaithyagr, line 101: You seem to be doing \Psi F*, the paper describes F* \psi right?

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
...
+
+        for i in range(s):
+            unit_vector[i] = 1
+            fourier_vector = nf.ifft(unit_vector)
+            coeffs = pw.wavedec(
+                fourier_vector, wavelet=wavelet_basis, level=nb_wavelet_scales
+            )
+            coeffs, _ = pw.coeffs_to_array(coeffs)
+            density_1d[i] = np.max(np.abs(coeffs)) ** 2
+            unit_vector[i] = 0
```
chaithyagr, line 126: Is it possible to vectorize this? How slow is it right now for 3D?

chaithyagr: @paquiteau , maybe numba?

src/mrinufft/trajectories/densities.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        coeffs, _ = pw.coeffs_to_array(coeffs)
+        density[ids] = np.max(np.abs(coeffs)) ** 2
+        unit_vector[ids] = 0
+
+    density = density / np.sum(density)
+    return nf.ifftshift(density)
+
+
+def create_fast_chauffert_density(shape, wavelet_basis, nb_wavelet_scales):
```
chaithyagr, line 110: To keep in mind: https://arxiv.org/pdf/1910.14513, some other interesting results with Anna.

chaithyagr (changes requested):
Starting my review...

chaithyagr:
> Based on this comment I'm not sure whether TSPx4 shown in panel (f) fulfills the equation because if you start from the TSD in (a) you should end up with the sqrt of (a) in (f). Fig. 6 in Nicolas' paper illustrates pretty well that phenomemon when contrasting panes (d) and (g).

I think the key missing ingridient here is this point by Guillaume:

> x4" (f) means that the trajectory consisted of a Travelling Salesman (TS) based on random sampling where only 1/4 of points were drawn, and 3/4 of points are simply interpolated along the TS path/solution.

I am assuming this means he is interpolating UNIFORMLY with 3 extra points in between every 2 points in trajectory. (@Daval-G to correct me).
Original work assumes that ALL the samples in between 2 points in TS solution are fully sampled (which results in the whole p^{-1/d} to begin with) rather than a fixed percentage. If we do that I am sure @Daval-G could have way more points which would potentially mess up his density. It is a different question about how to go about this. For now, it seems like as long as the trajectories are not scanner compliant, it really would not matter. But, if you plan to project this on constraint set, the density will drastically change.

src/mrinufft/trajectories/densities.py
Daval-G: It is true that those aliases are only used a couple of times, there is no need for them

src/mrinufft/trajectories/densities.py
Daval-G: Isn't `mrinufft/density` about density compensation ? Besides the name, there isn't much in common I think

src/mrinufft/trajectories/densities.py
Daval-G: Good point, I will do that

src/mrinufft/trajectories/densities.py
Daval-G: It is based on that equation actually. What difference do you see here ?

Daval-G: I am confused, this density is already based on that equation. What difference do you see ?
And what do you mean by elliptical sampling ? Anisotropic ? Why do you plan to disable it and from where?

src/mrinufft/trajectories/densities.py
Daval-G: I based this on the original code from Nicolas:
- https://github.com/philouc/mri_acq_recon_tutorial/blob/master/matlab/TSP_Markov/tools/compute_optimal_distrib.m
- https://github.com/philouc/mri_acq_recon_tutorial/blob/master/matlab/TSP_Markov/tools/perform_wavortho_transfAnis.m

I checked that the code had an identical behavior, and it is the case except for one major difference: he used hardcoded wavelets banks and I use `pywavelets` instead, and their values seem to differ.

src/mrinufft/trajectories/densities.py
Daval-G: Is there something specific in the paper you would like to highlight ? Should we open an issue to create some extension in the future ?

Daval-G: Also @chaithyagr I note in the paper figure that there is a TWIRL implementation sleeping somewhere in our cardboards. Do you know where it is so it can be added to the package ?

Daval-G:
> I am assuming this means he is interpolating UNIFORMLY with 3 extra points in between every 2 points in trajectory. (@Daval-G to correct me).

Actually I interpolate with cubic spline, it is maybe the best way to account for 1st and 2nd order derivative to get smooth transitions.

> Original work assumes that ALL the samples in between 2 points in TS solution are fully sampled (which results in the whole p^{-1/d} to begin with) rather than a fixed percentage. If we do that I am sure @Daval-G could have way more points which would potentially mess up his density.

That's a good point, so I carried an additional test where I just increase a lot the interpolation/oversampling while keeping the same number of TSP points (Ntsp). In the image below, we start with (b) at Nc=20 and Ns=200 but Ntsp=50, which is already 10% sampling/AF=10 compared to the 200x200 matrix size used for the density. The oversampling from Ntsp to Ns goes from 4 to 64 which is **_HUGE_**. I provide example trajectories below (clustered by bands to accelerate computations). At this point I think we can say that we sampled every point on the path followed by the TSP, and the histogram/observed density is still not affected.
oversampling_density

> For now, it seems like as long as the trajectories are not scanner compliant, it really would not matter. But, if you plan to project this on constraint set, the density will drastically change.

All the trajectories are defined over a normalized space [-0.5, 0.5]^d so there is necessarily a resolution and/or readout duration for which they are compliant. Also oversampling with cubic splines make transitions quite smooth already, given enough oversampling.

At this point I think it is relevant anyway to add an option to account for the `d/(d-1)` density exponent to fit the paper proposition.

Daval-G:
@chaithyagr I think the interpolation is indeed the key, I will carry extra tests this week to check that

chaithyagr:
Im sorry for the confusion, when I said UNIFORMLY in my above statement, I mean you added EXACTLY the same number of points in between any 2 points in the trajectory (which also makes sense).
But the original paper from what I see, seems to have more points if the points are far apart as compared to be near. I think we dont need to care about the extra factor for realistic usecases, but as you tell it is good to have an option, especially assuming that in the end we surely can get max gradient constrained making the overall sampling pattern different (and more close to whats shown in paper), but only upon projection.

src/mrinufft/trajectories/densities.py
chaithyagr: Isnt TWIRL just TPI in 2D? It should be fairly straightforward from our codes.

chaithyagr: > Is there something specific in the paper you would like to highlight ? Should we open an issue to create some extension in the future ?

I think @philouc can comment better than me. I just felt its good to add in docs, Its been long having gone through this work, but from what I remember its extension of the earlier work for anisotropic sampling, which can be of interest too.. But surely not to be pursued in this work

src/mrinufft/trajectories/densities.py
chaithyagr: Anisotropic and more importantly elliptical sampling, a way to enable or disable mask.

Daval-G: Somehow that dodges all of my questions xD Could you reformulate pls ?

chaithyagr: We need to support elliptical or non elliptical density

We also need to support anisotropy, but I think that we already do?

chaithyagr:
Great job on the experienent! Yes I agree with you on the default as False.

philouc:
Very nice job @Daval-G ; Your illustration in 2D is clear and very convincing. Additionally, suggestion about deactivating the density correction seems reasonable as the main purpose of this MRI-NUFFT package it to implement undersampling over a continuous k-space domain.

philouc:
Very sorry for the late feedback, I'm travelling every week!

src/mrinufft/trajectories/densities.py
Daval-G: Done

src/mrinufft/trajectories/densities.py
Daval-G: I had to fix it but I think it can now handle anisotropy and elliptical densities

Daval-G:
I will open an issue later to add `samplingsample_from_density` as an option for Wave-CAIPI rather than now

Daval-G:
@paquiteau I need to add dependencies to `pywavelets` (already mentioned) and `scikit-learn`. I see that we have a lot of optional dependency groups, which make sense for big GPU libraries that are not trivial to install, but among them I see really common libraries like `scikit-image` in `[smaps]`. Can't we just have those as base dependencies or do we really want nothing more than `numpy`, `scipy`, `matplotlib` & `tqdm` ?

paquiteau:
Indeed the dependencies management in MRI-Nufft could be improved. 
Aside the nufft backend, we could have an `extra` group where we have the dependencies you mentioned for trajectory as well as the `io` and `smaps` group. 

Then I think we could recommend people to `pip install mri-nufft[extras]` in the readme and docs

chaithyagr:
Maybe `scikit-learn` needs to be added for dependency in CI? I would recommend waiting for #195 to go in, which makes the process steramlined and only adding in pyproject.toml is enough

paquiteau:
the CI should simply install `mri-nufft[extras]` to reflect what user are doing :)

Daval-G:
@chaithyagr @paquiteau CI updated & passed. Let me know if good to merge, I am planning the next PRs already

paquiteau:
LGTM ! Thanks for this — once again — significant contribution :)

chaithyagr:
Given a lot of docs and also an example, can we run the docs in CI? You can get a run by adding `[docs]` tag in your commit.

chaithyagr:
Otherwise this PR LTGM..

PR #214, opened 2024-11-21
Fixes for normalization issues

### LINKED ISSUE #211: Inconsistent density-compensation normalization leading to signal instability
The normalization of density-compensation coefficients differs from the method used in the tfkbnufft package, resulting in signal energy instability when applying NUFFT with density compensation. 
We need to align the normalization approach with what is used in tfkbnufft.

AsmaTANABEN:
Adjust normalization of density compensation coefficients and adding Cufinufft Pipe function authored by Chaithya G.R.
Resolves #211


paquiteau (approved):
Ideally we have some tests as well ?

chaithyagr:
I would strongly recommend adding the changes for cufinufft in #195 as it is tested there. Also, that will allow this PR to be merged smoothly without waiting for finufft/cufinufft releases.

paquiteau:
Questions for you: can this normalization rule be used for the other density compensations methods (e.g. voronoi) ?

chaithyagr:
> Questions for you: can this normalization rule be used for the other density compensations methods (e.g. voronoi) ?

Answer for you: Yes :). It gives stable results.

chaithyagr (approved):
LGTM, assuming it matches same as tfkbnufft and tensorflow-nufft. I will rely on you for it @AsmaTANABEN , I dont think we need those tests here run regularly.

chaithyagr:
I think we can merge when green and @paquiteau lets make a release?

chaithyagr:
I think this can be merged right?

paquiteau:
We could merge it yes. However I think this normalization should be refactored to be generalizable to the other density compensation methods (e.g. voronoi, cell-counting etc.). If this is blocking we can post-pone this refactoring to another PR.

Issue #218, opened 2024-11-27
Fix a bunch of docs


chaithyagr:
This task tracks a bunch of doc fixes needed, most are quite short and small and fixes can be combined in a single PR:

- [x] #226
- [x] References in CG example https://mind-inria.github.io/mri-nufft/generated/autoexamples/GPU/example_cg.html needs to be made same as other example, i.e. at the end.
- [x] In CG method, make sure math is used for `Ax = B` in first cell.
- [x] https://mind-inria.github.io/mri-nufft/generated/autoexamples/GPU/example_learn_samples_multicoil.html : model_sense_op seems some minor fixes.
- [ ] Get utils.py somehow to binder/collab https://github.com/mind-inria/mri-nufft/blob/master/examples/utils.py

mcencini:
FYI: CG example has "Inpirations" instead of "Inspirations"

chaithyagr:
@alineyyy if you are touching the examples in #249 , can you handle the first 2 doc issues from here as well please?

PR #219, opened 2024-11-28
Update some examples. 

chaithyagr:
This is fairly straightforward minor changes and should not affect any of the CI

### DIFF
--- a/docs/binder/environment.yml
+++ b/docs/binder/environment.yml
@@ -9,4 +9,4 @@ dependencies:
       - mri-nufft[finufft]
       - joblib
       - brainweb-dl
-      - pytorch
+      - torch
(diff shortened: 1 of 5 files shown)


chaithyagr:
@paquiteau I dont think we need to wait for the tests to complete. Ill let you merge when you think the changes are good. I didnt want to add all these changes as a hotfix, these changes are just based on the issues I saw in docs. We need someway to check if the examples run fine on google-collab or binder, which is missing at the moment.

chaithyagr:
Also, note that examples using utils.py: https://github.com/mind-inria/mri-nufft/blob/fix_examples/examples/utils.py, would not work as the binder / collab instance does not have this file. We need to see how to fix that.. I have documented this in #218

paquiteau:
Do you have any idea on how you would automate the test for google collab ? (If we wish to do so, I am not so sure we should)

Otherwise, LGTM :)

chaithyagr:
> Do you have any idea on how you would automate the test for google collab ? (If we wish to do so, I am not so sure we should)
> 
> Otherwise, LGTM :)

No ideas xD

PR #220, opened 2024-12-02
Random sampling

paquiteau:
This PR adds random sampling extension for trajectories. 
In particular, it adds a trajectory expansion function (`stack_random`) to stack 2D trajectory following a certain distribution.

- [x] Add examples
- ~~([ ] Add ND sampling (e.g poisson disc))~~

An example of stack of spiral with uniform sampling (but fully sampled in the center of k-space)
Screenshot From 2024-12-02 15-15-50

### DIFF
--- a/examples/example_trajectory_tools.py
+++ b/examples/example_trajectory_tools.py
@@ -273,6 +273,44 @@ show_trajectories(
     axes=(1, 2),
 )
 
+# %%
+# Random Stacks
+# -------------
+#
+# A direct extension of the stacking expansion is to distribute the stacks
+# according to a random distribution over the :math:`k_z`-axis.
+#
+# Arguments:
+# -
+# ``trajectory (array)``: array of k-space coordinates of
+#  size :math:`(N_c, N_s, N_d)`
+# - ``dim_size (int)``: size of the kspace in voxel units
+# - ``center_prop  (int or float)`` : number of line
+# - ``acceleration (int)``:  Acceleration factor
+# - ``pdf (str or array)``: Probability density function for the random distribution
+# - ``rng (int or np.random.Generator)``: Random number generator
+# - ``order (int)``: Order of the shots in the stack
+
+
+trajectory = tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=128,
+    center_prop=0.1,
+    acceleration=4,
+    pdf="uniform",
+    order="center-out",
+)
+
+show_trajectory(trajectory, figure_size=figure_size, one_shot=one_shot)
+
+# %%
+# ``trajectory (array)``
+# ~~~~~~~~~~~~~~~~~~~~~~
+# The main use case is to stack trajectories consisting of
+# flat or thick planes that will match the image slices.
+#
+#
+#
 
 # %%
 # Precess
--- a/examples/example_trajectory_tools.py
+++ b/examples/example_trajectory_tools.py
@@ -219,6 +219,159 @@ show_trajectories(
     axes=(0, 2),
 )
 
+# %%
+# Stack Random
+# -------------
+#
+# A direct extension of the stacking expansion is to distribute the stacks
+# according to a random distribution over the :math:`k_z`-axis.
+#
+# Arguments:
+# - ``trajectory (array)``: array of k-space coordinates of size
+# :math:`(N_c, N_s, N_d)`
+# - ``dim_size (int)``: size of the kspace in voxel units
+# - ``center_prop  (int or float)`` : number of line
+# - ``acceleration (int)``:  Acceleration factor
+# - ``pdf (str or array)``: Probability density function for the random distribution
+# - ``rng (int or np.random.Generator)``: Random number generator
+# - ``order (int)``: Order of the shots in the stack
+
+
+trajectory = tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=128,
+    center_prop=0.1,
+    accel=16,
+    pdf="uniform",
+    order="top-down",
+    rng=42,
+)
+
+show_trajectory(trajectory, figure_size=figure_size, one_shot=one_shot)
+
+# %%
+# ``trajectory (array)``
+# ~~~~~~~~~~~~~~~~~~~~~~
+# The main use case is to stack trajectories consisting of
+# flat or thick planes that will match the image slices.
+arguments = ["Radial", "Spiral", "2D Cones", "3D Cones"]
+function = lambda x: tools.stack_random(
+    planar_trajectories[x],
+    dim_size=128,
+    center_prop=0.1,
+    accel=16,
+    pdf="gaussian",
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+# %%
+# ``dim_size (int)``
+# ~~~~~~~~~~~~~~~~~~
+# Size of the k-space in voxel units over the stacking direction. It
+# is used to normalize the stack positions, and is used with the ``accel``
+# factor and ``center_prop`` to determine the number of stacks.
+arguments = [32, 64, 128]
+function = lambda x: tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=x,
+    center_prop=0.1,
+    accel=8,
+    pdf="gaussian",
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+# %%
+# ``center_prop (int or float)``
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Number of lines to keep in the center of the k-space. It is used to determine
+# the number of stacks and the acceleration factor, and to keep the center of
+# the k-space with a higher density of shots. If a ``float`` this is a fraction
+# of the total ``dim_size``. If ``int`` it is directly the number of lines.
+
+arguments = [1, 5, 0.1, 0.5]
+function = lambda x: tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=128,
+    center_prop=x,
+    accel=16,
+    pdf="uniform",
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+
+# %%
+# ``accel (int)``
+# ~~~~~~~~~~~~~~~
+# Acceleration factor to subsample the outer region of the k-space.
+# Note that the acceleration factor does not take into account the center lines.
+
+
+arguments = [1, 4, 8, 16, 32]
+function = lambda x: tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=128,
+    center_prop=0.1,
+    accel=x,
+    pdf="uniform",
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+# %%
+# ``pdf (str or array)``
+# ~~~~~~~~~~~~~~~~~~~~~~
+# Probability density function for the sampling of the outer region. It can
+# either be a string to use a known probability law ("gaussian" or "uniform") or
+# "equispaced" for a coherent undersampling (like the one used in GRAPPA). It
+# can also be a array, for using a customed density probability.
+# In this case, it will be normalized so that ``sum(pdf) =1``.
+
+dim_size = 128
+arguments = [
+    "gaussian",
+    "uniform",
+    "equispaced",
+    np.arange(dim_size),
+]
+function = lambda x: tools.stack_random(
+    planar_trajectories["Spiral"],
+    dim_size=128,
+    center_prop=0.1,
+    accel=32,
+    pdf=x,
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+# %%
+# ``order (str)``
+# ~~~~~~~~~~~~~~~
+# Determine the ordering of the shot in the trajectory.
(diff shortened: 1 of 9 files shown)


src/mrinufft/trajectories/display.py
```diff
@@ -53,6 +53,8 @@ class displayConfig:
     This can be any of the matplotlib colormaps, or a list of colors."""
     one_shot_color: str = "k"
     """Matplotlib color for the highlighted shot, by default ``"k"`` (black)."""
+    one_shot_linewidth_factor: float = 2
+    """Factor to multiply the linewidth of the highlighted shot, by default ``2``."""
```
Daval-G, line 57: Sneaky but good point

src/mrinufft/trajectories/utils.py
```diff
@@ -144,6 +150,27 @@ class Packings(str, Enum, metaclass=CaseInsensitiveEnumMeta):
...
+# Variable Density Sampling #
+#############################
+
+
+class VDSorder(StrEnum):
+    """Available ordering for variable density sampling."""
+
+    CENTER_OUT = "center-out"
+    RANDOM = "random"
+    TOP_DOWN = "top-down"
```
Daval-G, line 163: It could be great at some point (not here) to have tools to manipulate the shot order

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+
+def get_random_loc_nd(N_samples, shape, pdf):
+    """Get random location in N-dimensions."""
+    # TODO
+    pass
+
+
+def stack_random(
+    traj,
+    dim_size,
```
Daval-G, line 837: Could you follow the naming conventions from this file ? (particularly `stack`)

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    center_prop,
+    accel=4,
+    pdf="uniform",
+    rng=None,
+    order="center-out",
```
Daval-G, line 842: As mentioned in another comment, it would be nice to (later) have ordering tools independently of those functions as an additional step. Similarly, some trajectories currently have an oversampling internally defined while it should just always be a step after the trajectory initialization through `tools.oversample` added in PR #206. WDYT ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    pass
+
+
+def stack_random(
+    traj,
+    dim_size,
+    center_prop,
+    accel=4,
+    pdf="uniform",
+    rng=None,
```
Daval-G, line 841: Having `rng` is a good point, should I add it also to PR #206 ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    """Stack a 2D trajectory with random location.
+
+    Parameters
+    ----------
+    traj: np.ndarray
+        Existing 2D trajectory.
+    dim_size: int
+        Size of the k_z dimension
+    center_prop: int or float
+        Number of line or proportion of slice to sample in the center of the k-space
```
Daval-G, line 853: Can you detail in the docstring the difference between `int` and `float` for `center_prop` ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    -------
+    numpy.ndarray
+        The 3D trajectory stacked along the :math:`k_z` axis.
+
+    """
+    line_locs = get_random_loc_1d(dim_size, center_prop, accel, pdf, rng, order)
+    if len(traj.shape) == 2:
+        Nc, Ns = 1, traj.shape[0]
+    else:
+        Nc, Ns = traj.shape[:2]
```
Daval-G, line 871:
```suggestion
    Nc, Ns = 1, traj.shape[0] if len(traj.shape) == 2 else traj.shape[:2]
```

Daval-G: Also the case where len(traj.shape) == 2 has never been considered until now. Is it something we expect to do ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+
+    new_traj = np.zeros((len(line_locs), Nc, Ns, 3))
+    for i, loc in enumerate(line_locs):
+        new_traj[i, :, :, :2] = traj
+        new_traj[i, :, :, 2] = loc
```
Daval-G, line 876: `tools.stack` and most other tools allow the input trajectory to also be 3D in case we want to stack "flat" trajectories with thickness as shown in the tools example in the documentation. Would it be possible to make that possible too for this function ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    order="center-out",
+):
+    """Stack a 2D trajectory with random location.
+
+    Parameters
+    ----------
+    traj: np.ndarray
+        Existing 2D trajectory.
+    dim_size: int
+        Size of the k_z dimension
```
Daval-G, line 851: Similarly to a comment above, please align the docstring on the existing ones from this file

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    center_prop: int or float
+        Number of line or proportion of slice to sample in the center of the k-space
+    pdf: str or np.array
+        Probability density function for the remaining samples.
+        "uniform" (default), "gaussian" or np.array
+    rng: random state
```
Daval-G, line 857: Type and definition ?

src/mrinufft/trajectories/tools.py
Daval-G, line 851: Does it imply that the stack is only on discrete space ?

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    """Get random location in N-dimensions."""
+    # TODO
+    pass
+
+
+def stack_random(
+    traj,
+    dim_size,
+    center_prop,
+    accel=4,
```
Daval-G, line 839: `accel` isn't defined in the docstring below. Also it is quite un-intuitive to use the term acceleration when we are adding extra shots rather than removing existing ones. In this file the convention is rather to give a multiplicative factor, as in `tools.stack` for example. I understand that `accel` comes along with `dim_size` but it is mostly to keep the draws over a discrete grid which I question too

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    Returns
+    -------
+    np.ndarray: array of size dim_size/accel.
+    """
+    order = VDSorder(order)
+    pdf = VDSpdf(pdf)
+    if accel == 0:
+        return np.arange(dim_size)  # type: ignore
+
+    indexes = list(range(dim_size))
```
Daval-G, line 771: Why pick positions over a discrete grid when uniform and gaussian pdfs are defined over continuous spaces ? We could remove `dim_size` entirely as it would be only relevant when `pdf` is an array which already contains `dim_size` in its shape

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    rng: random state
+
+    Returns
+    -------
+    np.ndarray: array of size dim_size/accel.
+    """
+    order = VDSorder(order)
+    pdf = VDSpdf(pdf)
+    if accel == 0:
+        return np.arange(dim_size)  # type: ignore
```
Daval-G, line 769: I am not sure why we would accept accelerations inferior or equal to 1

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+        line_locs = np.array(sorted(line_locs))
+    else:
+        raise ValueError(f"Unknown direction '{order}'.")
+    return (line_locs / dim_size) - 0.5  # rescale to [-0.5,0.5]
+
+
+def get_random_loc_nd(N_samples, shape, pdf):
+    """Get random location in N-dimensions."""
+    # TODO
+    pass
```
Daval-G, line 832: Isn't it basically the same as `sample_from_density` from PR #206 with ordering ? I would just separate sampling & ordering as mentioned in other comments

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    # apply order of lines
+    if order == VDSorder.CENTER_OUT:
+        line_locs = _flip2center(sorted(line_locs), dim_size // 2)
+    elif order == VDSorder.RANDOM:
+        line_locs = rng.permutation(line_locs)
+    elif order == VDSorder.TOP_DOWN:
+        line_locs = np.array(sorted(line_locs))
+    else:
+        raise ValueError(f"Unknown direction '{order}'.")
+    return (line_locs / dim_size) - 0.5  # rescale to [-0.5,0.5]
```
Daval-G, line 826: Please use KMAX instead of 0.5

src/mrinufft/trajectories/tools.py
```diff
@@ -700,3 +701,178 @@ def radialize_center(trajectory, nb_samples, in_out=False):
...
+    ----------
+    dim_size: int
+        Dimension size
+    center_prop: float or int
+        Proportion of center of kspace to continuouly sample
+    accel: float
+        Undersampling/Acceleration factor
+    pdf: str, optional
+        Probability density function for the remaining samples.
+        "gaussian" (default) or "uniform".
```
Daval-G, line 759: In `stack_random` below, this argument can also be an array. I don't think this function can handle it, does it ?

Daval-G (changes requested):
Great idea for an additional tool. A bunch of comments mostly about aligning this function with other similar tools and also avoiding redundancy with PR #206

src/mrinufft/trajectories/tools.py
paquiteau: Yes, its a sane default when coupled with `rng=np.random.default_rng(rng)` for validation

src/mrinufft/trajectories/tools.py
paquiteau: That's a good point. There are two reason: 1) this function was initially written 2 years ago for creating cartesian mask, and it evolved from there 2) this enforces a stacking that lies on the cartesian grid, making it suitable for Stacked-NUFFT (2.5 D) optimization. 

A solution could be to resort to the continuous case if `dim_size = None` ?

src/mrinufft/trajectories/tools.py
paquiteau: accel = 1 would get you a fully sampled trajectory along one axis, basically making `stack_random` like a variant of `stack`

src/mrinufft/trajectories/tools.py
paquiteau: having a single oversampling function can be nice indeed.

src/mrinufft/trajectories/tools.py
paquiteau: it does.

src/mrinufft/trajectories/tools.py
paquiteau: I did not find a good way to do that, as the number of stacks is determined from multiples parameters (accel, and acs)

paquiteau:
@Daval-G What do you think of this updated version ? also do not hesitate to check the example as well

Daval-G:
@paquiteau I am starting again to look at this request, but I won't finish tonight. I see that a bunch of my old comments are un-answered, could you have a look ? Thx

paquiteau:
I think now I have address everything you asked for :)

src/mrinufft/trajectories/tools.py
```diff
@@ -784,3 +785,180 @@ def radialize_center(
...
+    center_indexes = indexes[center_start:center_stop]
+    borders = np.asarray([*indexes[:center_start], *indexes[center_stop:]])
+
+    n_samples_borders = (dim_size - len(center_indexes)) // accel
+    if n_samples_borders < 1:
+        raise ValueError(
+            "acceleration factor, center_prop and dimension not compatible."
+            "Edges will not be sampled. "
+        )
+    rng = np.random.default_rng(rng)  # validate get a RNG from a seed or existing rng.
```
Daval-G, line 875: What does the comment mean here ?

Daval-G (changes requested):
Great improvements, only small comments remaining. Would you like to give a try to type hints ? If not it's OK, it was an itch from me but I don't want to impose it

examples/example_trajectory_tools.py
```diff
@@ -273,6 +273,161 @@
...
+    pdf="gaussian",
+    order="top-down",
+    rng=42,
+)
+show_trajectories(function, arguments, one_shot=one_shot, subfig_size=subfigure_size)
+
+# %%
+# ``dim_size (int)``
+# ~~~~~~~~~~~~~~~~~~
+# Size of the k-space in voxel units. It
```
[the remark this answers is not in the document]
paquiteau: yep that could be confusing, fixed it :)

Daval-G (approved):
LGTM

PR #224, opened 2024-12-06
Add support for fiNUFFT spread interpolate

chaithyagr:
This handles the final piece in puzzle, i.e. density compensation on CPU

This depends on #195 to go in and later https://github.com/flatironinstitute/finufft/pull/599 to be done and a release up.


src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -87,13 +85,15 @@ def _make_plan(self, typ, **kwargs):
...
+            self._kz.set(samples[:, 2])
+
+    def _set_pts(self, typ):
         plan = self.grad_plan if typ == "grad" else self.plans[typ]
-        plan.setpts(
-            cp.array(samples[:, 0], copy=False),
-            cp.array(samples[:, 1], copy=False),
-            cp.array(samples[:, 2], copy=False) if self.ndim == 3 else None,
-        )
+        plan.setpts(self._kx, self._ky, self._kz)
```
paquiteau, line 96: Seems redundant?

src/mrinufft/operators/interfaces/utils/utils.py
```diff
@@ -28,3 +26,29 @@ def sizeof_fmt(num, suffix="B"):
...
+    while numdiv > 1 or nplus % b != 0:
+        nplus += 2  # stays even
+        numdiv = nplus
+        while numdiv % 2 == 0:
+            numdiv //= 2  # remove all factors of 2, 3, 5...
+        while numdiv % 3 == 0:
+            numdiv //= 3
+        while numdiv % 5 == 0:
+            numdiv //= 5
+    return nplus
```
paquiteau, line 54: I did not find any use of this function elsewhere. Is it really necessary ?

.github/workflows/test-ci.yml
```diff
@@ -92,13 +65,6 @@ jobs:
           make
           echo $PWD >> $GITHUB_PATH
 
-      - name: Install torchkbnufft-cpu
-        if: ${{ matrix.backend == 'torchkbnufft-cpu'  || env.ref_backend == 'torchkbnufft-cpu'}}
-        run: |
-          ${{ env.activate_venv }}
-          python -m pip install torchkbnufft
```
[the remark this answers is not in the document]
chaithyagr: Well I anticipated this:
```
torchkbnufft = ["torchkbnufft", "cupy-cuda12x"]
torchkbnufft-cpu = ["torchkbnufft", "cupy-cuda12x"]
torchkbnufft-gpu = ["torchkbnufft", "cupy-cuda12x"]
```

Seeeee :P

src/mrinufft/operators/interfaces/utils/utils.py
chaithyagr: Good catch, it was old obselete codes

src/mrinufft/operators/interfaces/cufinufft.py
chaithyagr: Recommendation to have it better?

.github/workflows/test-ci.yml
```diff
@@ -127,33 +93,22 @@ jobs:
...
-          elif [[ ${{ matrix.backend }} == "cufinufft" ]]; then
-            python -m pip install "cufinufft<2.3"
-          else
-            python -m pip install ${{ matrix.backend }}
-          fi
+          source $RUNNER_WORKSPACE/venv/bin/activate
+          export CUDA_BIN_PATH=/usr/local/cuda-12.4/
+          export PATH=/usr/local/cuda-12.4/bin/:${PATH}
+          export LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64/:${LD_LIBRARY_PATH}
+          pip install -e .[${{ matrix.backend }},autodiff]
```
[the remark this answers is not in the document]
chaithyagr: Same as above :)

.github/workflows/test-ci.yml
```diff
@@ -377,23 +337,22 @@ jobs:
...
         with:
           python-version: "3.10"
-
+        
+      - name: Point to CUDA 12.4
+        run: |
+          export CUDA_BIN_PATH=/usr/local/cuda-12.4/
+          export PATH=/usr/local/cuda-12.4/bin/:${PATH}
+          export LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64/:${LD_LIBRARY_PATH}
+
```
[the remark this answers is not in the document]
chaithyagr: I dont think so? The tests would fail otherwise in CI I'm guessing. Atleast the installation of gpuNUFFT

paquiteau:
not quite a fan of the reduced accuracy for finufft, what changed it ? Also the CI reports strong mismatch. so the problem may go deeper...

chaithyagr:
We need to do something about the tests being very strict, here is the actual images plots where we have maximum difference in output and input:

Most tests expects the error to be less than 1e-3 of absolute error 
For example for test_batch_adj_op:

Overlap of 
image

Errors:
image

Zoom in of errors:

image



For now, I reduced most of the strictness, but only for finufft. We can open an issue to investigate what is happening, especially as it is happening only on CI server! Locally everything is running fine!

paquiteau:
What is the reference here ? maybe the "error" comes from the reference, and not finufft on the CI server ?

chaithyagr:
Its passing!!!!!! @paquiteau its passing!!!

src/mrinufft/operators/interfaces/cufinufft.py
paquiteau: well I don't understand why you need `_set_pts` and `_set_kxyz` for me `_set_pts(self, typ, samples)` should be enough. keeping an attribute `self._samples` would also be simpler (IMO) than `_kx,_ky,_kz`.

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -850,3 +850,55 @@ def toggle_grad_traj(self):
...
+            The oversampling factor the volume shape
+        normalize: bool
+            Whether to normalize the density compensation.
+            We normalize such that the energy of PSF = 1
+        """
+        if CUFINUFFT_AVAILABLE is False:
+            raise ValueError(
+                "cufinufft is not available, cannot estimate the density compensation"
+            )
+        grid_op = MRICufiNUFFT(
```
paquiteau, line 884: Since this is a classmethod, use `cls` instead of the backend name. Ultimately this could be refactorized between all backend that supports "pipe" density (with an extra attention to the cupy/numpy array interface)

src/mrinufft/operators/interfaces/finufft.py
```diff
@@ -168,3 +168,55 @@ def toggle_grad_traj(self):
...
+                grid_op.op(
+                    grid_op.adj_op(density_comp.astype(grid_op.cpx_dtype))
+                ).squeeze()
+            )
+        if normalize:
+            test_op = MRIfinufft(samples=kspace_loc, shape=volume_shape, **kwargs)
+            test_im = np.ones(volume_shape, dtype=test_op.cpx_dtype)
+            test_im_recon = test_op.adj_op(density_comp * test_op.op(test_im))
+            density_comp /= np.mean(np.abs(test_im_recon))
+        return density_comp.squeeze()
```
paquiteau, line 222: same comment as for cufinufft, use `cls`, maybe consider refactoring

paquiteau (changes requested):
Some nitpicks, but that's very good :) Thank you for the hard work !

src/mrinufft/operators/interfaces/finufft.py
chaithyagr: Ill use cls, but what do you mean by refactoring?

paquiteau: have a single pipe function in the code base (and not 3 different implementation for finufft, cufinufft, gpunufft). Maybe a mixin class could do the trick.

chaithyagr: Ahh yes, handling it in #231 for it would be ideal.. i don't want to burden this PR

paquiteau (approved):
LGTM

Let's keep in mind this accuracy issue for later

chaithyagr:
Hey @kborsos , feel free to use `pip install git+https://github.com/mind-inria/mri-nufft#egg=mri-nufft[cufinufft]` to already use it. We are waiting a release from the cufinufft folks (which is currently a pre-release) to make our release, to avoid issues for users. 
However, the above command works just fine!

PR #228, opened 2024-12-18
Clean documentation and add type hints

Daval-G:
Just cleaning the documentation a bit and adding type hints

### DIFF
--- a/src/mrinufft/trajectories/display.py
+++ b/src/mrinufft/trajectories/display.py
@@ -164,7 +164,7 @@ def _setup_3D_ticks(figsize: float, fig: plt.Figure | None = None) -> plt.Axes:
 
 
 def display_2D_trajectory(
-    trajectory: np.ndarray,
+    trajectory: np.typing.NDArray,
     figsize: float = 5,
     one_shot: bool | int = False,
     subfigure: plt.Figure | plt.Axes | None = None,
@@ -279,7 +279,7 @@ def display_2D_trajectory(
 
 
 def display_3D_trajectory(
-    trajectory: np.ndarray,
+    trajectory: np.typing.NDArray,
     nb_repetitions: int | None = None,
     figsize: float = 5,
     per_plane: bool = True,
@@ -418,7 +418,7 @@ def display_3D_trajectory(
 
 
 def display_gradients_simply(
-    trajectory: np.ndarray,
+    trajectory: np.typing.NDArray,
     shot_ids: tuple[int, ...] = (0,),
     figsize: float = 5,
     fill_area: bool = True,
@@ -532,7 +532,7 @@ def display_gradients_simply(
 
 
 def display_gradients(
-    trajectory: np.ndarray,
+    trajectory: np.typing.NDArray,
     shot_ids: tuple[int, ...] = (0,),
     figsize: float = 5,
     fill_area: bool = True,
--- a/src/mrinufft/trajectories/display.py
+++ b/src/mrinufft/trajectories/display.py
@@ -47,7 +47,7 @@ class displayConfig:
     """Font size for most labels and texts, by default ``18``."""
     small_fontsize: int = 14
     """Font size for smaller texts, by default ``14``."""
-    nb_colors = 10
+    nb_colors: int = 10
     """Number of colors to use in the color cycle, by default ``10``."""
     palette: str = "tab10"
     """Name of the color palette to use, by default ``"tab10"``.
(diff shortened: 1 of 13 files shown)


paquiteau:
I am not against typing, however we should do it properly have you considered using mypy  and /or ruff (the `ANN*` errors are disabled for now). 

The trajectory module is a great starting point for its usage, and would likely complement the docstring to help people play with them. 

A few comments:

- try to avoid `Any` 
- `str` type are most of the time limited to named values , you can use `Literal` instead 
- `*args` and `**kwargs` should be typed with the type of the elements, not `list`

Daval-G:
Good point, I will do that.

Daval-G:
Good point. I followed ruff's ANN rule, and the current code is compliant (except for bad usage of Literal apparently)

> try to avoid Any

Done, but some cases like open kwargs are still relevant. I tried to reduce them to a minimum though.

> str type are most of the time limited to named values , you can use Literal instead

Make sense. The CI tells me I underestimated `Literal`, I will fix that soon.

> *args and **kwargs should be typed with the type of the elements, not list

Make sense.

> typing np array is not as good as it could be the best future prood solution is to use np.typing.NDArray

Done, but I have to check also how it renders in docstrings

src/mrinufft/trajectories/display.py
```diff
@@ -58,33 +59,33 @@ class displayConfig:
...
 
-    def reset(self):
+    def reset(self) -> None:
         """Restore the display configuration."""
         for key, value in self._old_values.items():
             setattr(displayConfig, key, value)
         delattr(self, "_old_values")
 
-    def __enter__(self):
+    def __enter__(self) -> "displayConfig":
```
paquiteau, line 79: if you use `from __future__ import annotations` you should be able to annotate with the Class directly, not a string :)

src/mrinufft/trajectories/display.py
```diff
@@ -163,21 +164,21 @@ def _setup_3D_ticks(figsize, fig=None):
...
-    figsize=5,
-    one_shot=False,
-    subfigure=None,
-    show_constraints=False,
-    gmax=DEFAULT_GMAX,
-    smax=DEFAULT_SMAX,
-    constraints_order=None,
-    **constraints_kwargs,
-):
+    trajectory: np.typing.NDArray,
```
paquiteau, line 167: I think you can just `from numpy.typing import NDArray` and use it everywhere

src/mrinufft/trajectories/inits/random_walk.py
```diff
@@ -20,8 +22,14 @@ def _get_neighbors_offsets(shape):
...
 def _initialize_ND_random_walk(
-    Nc, Ns, density, *, diagonals=True, pseudo_random=True, **sampling_kwargs
-):
+    Nc: int,
+    Ns: int,
+    density: np.typing.NDArray,
+    *,
+    diagonals: bool = True,
+    pseudo_random: bool = True,
+    **sampling_kwargs: Literal | bool,
```
paquiteau, line 31: `Literal` has to be used as  `Literal["A string"]` or `Literal[1]` for instance

src/mrinufft/trajectories/inits/random_walk.py
```diff
@@ -88,8 +95,14 @@ def _initialize_ND_random_walk(
...
+) -> np.ndarray:
```
paquiteau, line 105: should be `NDArray`

paquiteau (changes requested):
We are on a good path, thanks for the tedious work. 

Do not hesitate to check your modification with mypy locally. I am also pretty sure an LLM could help you adding annotation in bulk if needed...

src/mrinufft/trajectories/display.py
Daval-G: Delayed annotation checks helped in several cases actually, thx !

src/mrinufft/trajectories/display.py
Daval-G: Done

src/mrinufft/trajectories/inits/random_walk.py
Daval-G: Done. Most `Literal` uses were actually unjustified because of the way we handle `Enum` for everything

src/mrinufft/trajectories/inits/random_walk.py
Daval-G: Yes I missed a lot of them for some reason, thx

Daval-G:
mypy is still showing some issues but I reduced them from 120 to 25, with half of them outside the trajectory submodule for some reason. The current code is compliant with ruff's ANN rules

src/mrinufft/trajectories/inits/travelling_salesman.py
```diff
@@ -95,15 +98,15 @@ def _cluster_by_coordinate(
...
     Nc: int,
     Ns: int,
-    density: np.typing.NDArray,
-    first_cluster_by: Literal | None = None,
-    second_cluster_by: Literal | None = None,
-    sort_by: Literal | None = None,
+    density: NDArray,
+    first_cluster_by: Literal["x", "y", "z", "r", "phi", "theta"] | None = None,
+    second_cluster_by: Literal["x", "y", "z", "r", "phi", "theta"] | None = None,
+    sort_by: Literal["x", "y", "z", "r", "phi", "theta"] | None = None,
```
paquiteau, line 104: I think you can:

- Use a type alias (declared below the imports), for brievety
- Put the None inside the literal

PR #230, opened 2025-01-09
Finally get right Orientation from raw data

chaithyagr:
This is continuation of saga in #157 which was based on twixtools, which was not working well to read data. 
Finally we have this working.

Some notes for how I got this working:

The base notes of how quaternion is ordered in data: https://github.com/ismrmrd/siemens_to_ismrmrd/blob/993ff5ffa74b4f719af36264e98698b05683498b/main.cpp#L1131-L1142

Which calls : 
https://github.com/ismrmrd/ismrmrd/blob/802b52d69dcc31935888d3c69cc2a179f13a5608/libsrc/ismrmrd.c#L668-L683

However notice that call is Phase Read Slice, while the function expects Read Phase Slice. This mis-match is as data is expected to be Y X Z co-ordinates, cause I dont know why (@paquiteau has some links xD) 

Based on this I almost got it working except sometimes the brain was inverted, and I found: https://github.com/ismrmrd/ismrmrd/blob/802b52d69dcc31935888d3c69cc2a179f13a5608/libsrc/ismrmrd.c#L578-L610

Where you try to change orientation based on if determinant of rotation matrix is negative (to keep it 1)


src/mrinufft/io/utils.py
```diff
@@ -35,3 +35,34 @@ def add_phase_to_kspace_with_shifts(kspace_data, kspace_loc, normalized_shifts):
...
+    R[1,2] = 2.0 * (b * c - a * d)
+
+    R[2,1] = 2.0 * (a * c - b * d)
+    R[2,0] = 2.0 * (b * c + a * d)
+    R[2,2] = 1.0 - 2.0 * (a * a + b * b)
+
+    if (np.linalg.det(R[:3, :3]) < 0):
+        R[2] = -R[2]
+        
+    R[-1,-1] = 1
```
paquiteau, line 66: Why do it here rather than before the determinant check ?

paquiteau (commented):
Nice job ! 

How do you want to use the orientation attribute later on ? 

also, it may be nice to use: 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html

for the computation (that way I won't ask for a unit test :P )

src/mrinufft/io/utils.py
chaithyagr: Saving compute xD. determinant of 3x3 is faster i presume than 4x4 :P

chaithyagr:
> also, it may be nice to use:
> 
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html
> 
> for the computation (that way I won't ask for a unit test :P )

I really wish I could, I tried a lot and could not get the 2 to match ...

chaithyagr:
I think its because of all the intricacies like having it as Y X Z coordinates etc. Lemme see if I can fix it though

PR #233, opened 2025-02-04
Add some extra features to Siemens reader.

chaithyagr:
We add suppoort to read ACS region (useful for GRAPPA + GoLF and maybe in future EPI)


src/mrinufft/io/utils.py
```diff
@@ -60,3 +60,28 @@ def siemens_quat_to_rot_mat(quat):
...
     R[-1, -1] = 1
     return R
+
+
+def remove_extra_kspace_samples(kspace_data, num_samples_per_shot):
+    """Remove extra samples from k-space data.
+
+    This function is useful when the k-space data has extra samples
+    mainly as ADC samples at only upto
+
```
paquiteau, line 70: finish the sentence ?

paquiteau (changes requested):
Seems reasonable to do for your application. Also, be sure to update the documentation about the extra header field.

src/mrinufft/io/utils.py
chaithyagr: Lol this is what happens when someone calls me to show their fancy results while writing code :P

chaithyagr:
@paquiteau can we merge this?

paquiteau (approved):

PR #129, opened 2024-06-12
Vds spiral

paquiteau:
This PR adds support for Variable density spiral generation. 

It is a port to Python of the code of Brian Hargreaves from 

http://mrsrl.stanford.edu/~brian/vdspiral/

The associated publication is here: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.10644

Note that the units are messed up for now compared to the conventions of MRI-NUFFT, but this can still provide some good results.

On top of the original code, I added support for in-out trajectories.

This opens new potential for MRI-NUFFT as this is the first trajectory generation that is procedural and dependent on target FOV. It is an early development phase, and a few things I wish would be addressed in the near future:

- [ ] Fix In-out trajectories by correct initialization of the gradients 
- [ ] Use the correct unit and scaling (Telsa /meter/seconds / Hertz)
- [ ] Support density profiles that are not polynomial 
- [ ] The code is quite slow compared to Matlab; there might be some optimization to do. 

image
image

PR #236, opened 2025-02-14
Add arc length parameterization

Daval-G:
This PR adds a tool to minimize the gradient amplitudes using arc length parameterization. It makes the samples evenly distributed along the path of each shot while maintaining the full shot length, and therefore the maximum gradient amplitude is optimally minimized as all gradient strengths are the same (but not the gradient directions of course). The catch is that the maximum slew rate can become insanely high if the path is not smooth enough. An example is shown below.

Screenshot from 2025-02-15 00-41-06
(A) A classic (blue) spiral trajectory is projected (orange) to be uniformly distributed over the spiral path. (B) shows the gradient amplitudes and (C) shows the slew rates. The projected gradient strength is completely flat but it leads to much stronger slew rates around the center where the spiral direction changes.

### DIFF
--- a/src/mrinufft/trajectories/projection.py
+++ b/src/mrinufft/trajectories/projection.py
@@ -1,0 +1,58 @@
+"""Functions to fit gradient constraints."""
+
+import numpy as np
+from numpy.typing import NDArray
+from scipy.interpolate import CubicSpline
+
+
+def fit_arc_length(
+    trajectory: NDArray, order: int | None = None, eps: float = 1e-8
+) -> NDArray:
+    """Adjust the trajectory to have a uniform arc-length distribution.
+
+    The trajectory is parametrized according to its arc length along a
+    cubic-interpolated path and samples are repositioned to minimize
+    the gradients amplitude. This solution is optimal with respect to
+    gradients but can lead to excessive slew rates, and it will change
+    the overall density.
+
+    Parameters
+    ----------
+    trajectory: NDArray
+        A 2D or 3D trajectory of shape (Nc, Ns, Nd), with Nc the number of shots,
+        Ns the number of samples per shot, and Nd the number of dimensions.
+    order: int | None
+        The order of the norm used to compute arc length, based on the convention from
+        `numpy.linalg.norm`. Defaults to None (Euclidean norm).
+    eps: float
+        Convergence threshold for stopping the iterative refinement. Defaults to 1e-8.
+
+    Returns
+    -------
+    NDArray: The reparameterized trajectory with the same shape as the input.
+    """
+    Nc, Ns, Nd = trajectory.shape
+    new_trajectory = np.copy(trajectory)
+
+    for i in range(Nc):
+        time = np.linspace(0, 1, Ns)
+        projection = trajectory[i]
+        old_projection = 0
+        arc_func = CubicSpline(time, projection)
+
+        while (
+            np.linalg.norm(projection - old_projection) / np.linalg.norm(projection)
+            > eps
+        ):
+            arc_length = np.cumsum(
+                np.linalg.norm(np.diff(projection, axis=0), ord=order, axis=-1), axis=0
+            )
+            arc_length = np.concatenate([[0], arc_length])
+            arc_length = arc_length / arc_length[-1]
+            inv_arc_func = CubicSpline(arc_length, time)
+
+            time = inv_arc_func(np.linspace(0, 1, Ns))
+            old_projection = np.copy(projection)
+            projection = arc_func(time)
+        new_trajectory[i] = projection
+    return new_trajectory
--- a/src/mrinufft/trajectories/trajectory3D.py
+++ b/src/mrinufft/trajectories/trajectory3D.py
@@ -17,6 +17,7 @@ from .maths import (
     Rz,
     generate_fibonacci_circle,
 )
+from .projection import fit_arc_length
 from .tools import conify, duplicate_along_axes, epify, precess, stack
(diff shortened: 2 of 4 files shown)


Daval-G:
It appears that the `initialize_3D_annular_shells` function has zero-gradient transitions sometimes, making the interpolation fail. This should not be the case in any trajectory, so for now I am going to make the projection compatible with that case but next MR I will add generic unit tests to ensure none of the trajectories has that

paquiteau:
Nice stuff! I wonder if this can be useful (and maybe supersede) #129 ? Regarding the null gradient problem: From my understanding, it should be equivalent to have the same point sampled consequently multiple times. I guess typically this not desirable, but there might be cases where this could be considered?

chaithyagr:
Awesome stuff. In your short example picture, is it possible to have arc length only above some frequency in kspacr to ensure we don't have slew issues at center due to turning at high speeds

src/mrinufft/trajectories/projection.py
```diff
@@ -0,0 +1,60 @@
...
+
+def fit_arc_length(
+    trajectory: NDArray, order: int | None = None, eps: float = 1e-8
+) -> NDArray:
+    """Adjust the trajectory to have a uniform arc-length distribution.
+
+    The trajectory is parametrized according to its arc length along a
+    cubic-interpolated path and samples are repositioned to minimize
+    the gradients amplitude. This solution is optimal with respect to
+    gradients but can lead to excessive slew rates, and it will change
```
chaithyagr, line 17: Both these points are important and are worth adding as a note or warning

src/mrinufft/trajectories/projection.py
```diff
@@ -0,0 +1,60 @@
+"""Functions to fit gradient constraints."""
+
+import numpy as np
+import numpy.linalg as nl
+from numpy.typing import NDArray
+from scipy.interpolate import CubicSpline
+
+
+def fit_arc_length(
```
chaithyagr, line 9: Maybe call it fit_constant_arc_length ?

chaithyagr: Also it would be worth trying to add support for this parameterization only in certain portions of kspace , but i think it's not easy as you might have discontinuity

chaithyagr (changes requested):
Some first round of reviews... Minor comments for now n some to think about. Thank you for this!

src/mrinufft/trajectories/projection.py
Daval-G: > Maybe call it fit_constant_arc_length ?

Discussing with ChatGPT made me realize using `fit` isn't appropriate, WDYT ? I consider solutions like `resample_by_arc_length`, `make_uniform_arc_length`, `balance_arc_length`, `equalize_arc_length`

Daval-G: > Also it would be worth trying to add support for this parameterization only in certain portions of kspace , but i think it's not easy as you might have discontinuity

It would be the same as using the function but giving as input just a subpart of the trajectory. Would it be OK with you if I show explicitly how to do that in the example rather than changing the function for something the user can do?

chaithyagr: I am okay with that surely, but I am worried that we will not be able to handle the discontinuity well.. We potentially will have slew violations there/

src/mrinufft/trajectories/trajectory3D.py
```diff
@@ -728,6 +729,9 @@ def initialize_3D_annular_shells(
     """Initialize 3D trajectories with annular shells.
 
     An exclusive trajectory inspired from the work proposed in [HM11]_.
+    It is similar to other trajectories based on concentric rings but
+    rings are split and recombined by pairs to better balance
+    the shot lengths.
```
chaithyagr, line 734: I dont fully understand this... split and recombined by pairs? 

Also, can I say that earlier implementation was wrong as this function didn't exist, or did this function make this implementation better?

chaithyagr (changes requested):
Some more comments.. Adding example would be helpful. 
Also @paquiteau its time we start adding examples which is using this function feature at the end of docs. WDYT?

paquiteau:
Let's tackle the doc /example links in another PR (#237 for further discussion)

chaithyagr:
Just noting here that #272 should be able to handle this.

src/mrinufft/trajectories/trajectory3D.py
Daval-G: Updated, let me know if it's more clear :)

src/mrinufft/trajectories/projection.py
Daval-G: After trying it's actually not possible, or more precisely it's far beyond the scope of that method as it would imply working on the slew rate too, which is what #272 is doing more seriously

Daval-G:
@chaithyagr WDYT ?

src/mrinufft/trajectories/trajectory3D.py
```diff
@@ -832,6 +837,8 @@ def initialize_3D_annular_shells(
         trajectory[count : count + Ms] = shell
         count += Ms
 
+    # Balance the samples over the recombined shots
+    # trajectory = parameterize_by_arc_length(trajectory) # FIXME
```
chaithyagr, line 841: Can you please add a comment explaining the FIXME

chaithyagr (approved):
Just minor changes. Please update it and if possible open an issue to track it. Otherwise LGTM, just need tests passing. You can merge when green and after handeling this. I dont think we need another review. Thanks again for your wonderful work

Issue #244, opened 2025-03-03
Finufft Tests are very slow


paquiteau:
On the local runner instance it seems that the test using finufft has a backend are running forever (more than 1hour) 

to reproduce locally  I ran 
```
pytest --backend=finufft --ref=finufft --durations=10 --cov
```

NB: the `durations` arg is there to track how much time each test took

```
============================ slowest 10 durations ============================
69.51s call     tests/operators/test_offres_exp_approx.py::test_zmap_coeff[complex3D-numpy]
68.85s call     tests/operators/test_offres_exp_approx.py::test_b0map_coeff[real3D-numpy]
37.10s call     tests/operators/test_offres_exp_approx.py::test_zmap_coeff[complex3D-torch-cpu]
36.16s call     tests/operators/test_offres_exp_approx.py::test_b0map_coeff[real3D-torch-cpu]
17.34s call     tests/test_ndft.py::test_ndft_implicit1[grid3D]
13.86s call     tests/operators/test_offres_exp_approx.py::test_b0map_coeff[real3D-torch-gpu]
12.18s call     tests/operators/test_offres_exp_approx.py::test_b0map_coeff[real3D-cupy]
11.22s call     tests/operators/test_offres_exp_approx.py::test_zmap_coeff[complex3D-cupy]
9.25s call     tests/test_ndft.py::test_ndft_implicit2[grid3D]
7.65s call     tests/operators/test_interfaces.py::test_interface_lipschitz[random3D-finufft]
```

Overall, it took 10 minutes to run this command, way more reasonable than the 1h30 on the CI

paquiteau:
From offline discussion in order to make the CI even faster:
- [ ] Extract the `offres_ex_approx.py` tests to not be runned with operators (They don't call any nufft actually)
- [ ] filter test_examples (e.g with commit head line)

PR #245, opened 2025-03-03
Fix ci, round 2

paquiteau:
In this PR we want to make test run faster, solving #244


paquiteau:
from the CI runs, it seems that expensive tests are run multiple times (notably the off-resonnances one). It might be best to filter them and run them separately.

PR #246, opened 2025-03-11
Clean signal gradient display

Daval-G:
Minor PR to fix the "signal" display for gradient diagrams, that was needlessly complex and not even properly representing signal as intended.

Old display

New display


chaithyagr (approved):
LGTM

PR #247, opened 2025-03-17
Fix Lloyd's sampling coordinates

Daval-G:
This PR fixes the coordinates fitted by the Lloyd sampling algorithm. As shown in the image below, each voxel was shifted by half a voxel width which lead to errors when gridding the locations for example. The new fitted coordinates correspond to what was already done for the random sampling, making both behaviors match.

Screenshot from 2025-03-18 00-38-59


paquiteau (approved):

Daval-G:
Unrelated fail from the CI, one approval. I merge :D

Issue #254, opened 2025-04-09
Off-resonance correction works in opposite convention


chaithyagr:
Recently, we started to look into off-resonance correction codes in mri-nufft and we noticed that the off-resonance term seems to be with a `-` sign here, which means the correction occurs only when we invert the field map given as input.

This goes against our usual convention where we have a field map and we want to correct for artifacts. 
I don't know if this is opposite of the convention used by you folks @mcencini ? 

@Daval-G thoughts? 

Also, I noticed we don't have tests for it here. we can use this issue to track its addition here.

https://github.com/mind-inria/mri-nufft/blob/cbcdf2abfb61f97047d15f6de3d67af3c4e9708f/src/mrinufft/operators/off_resonance.py#L371

mcencini:
Hi!

I used MIRT matlab implementation (
https://github.com/JeffFessler/mirt/blob/main/mri/mri_exp_approx.m) as a
reference, also using - sign.

They also have GE, and IIRC GE scanners (at least ours) have opposite B0
sign. Maybe we put a flag to select the sign?

RE unit tests: I only tested for coefficients because I struggled to find a
clear test case for forward and adjoint methods. Now that I am thinking
about it, what about a homogeneous off resonance field with cartesian
readout? It should produce a quite predictable translation of the object.

Il mer 9 apr 2025, 12:31 Chaithya G R ***@***.***> ha scritto:

PR #256, opened 2025-04-17
 Fix : add negate flag to switch sign convention in ORC

femaleprog:
This PR adds a negate flag to allow switching the sign of the field map in off-resonance correction, addressing issue [#254](https://github.com/mind-inria/mri-nufft/issues/254).

Default remains backward-compatible (negate=True)

If negate=False, the phase factor uses a + sign (conventional)

### DIFF
--- a/src/mrinufft/operators/off_resonance.py
+++ b/src/mrinufft/operators/off_resonance.py
@@ -199,7 +199,7 @@ class MRIFourierCorrected(FourierOperatorBase):
         The default is ``None`` (purely imaginary field).
         Also supports Cupy arrays and Torch tensors.
     isign : int, optional
-        Sign convention for spatial phase. 
+        Sign convention for spatial phase.
         Either `-1` (default, e^{-iωt}) or `1` (e^{iωt}).
     backend: str, optional
         The backend to use for computations. Either 'cpu', 'gpu' or 'torch'.
--- a/src/mrinufft/operators/off_resonance.py
+++ b/src/mrinufft/operators/off_resonance.py
@@ -269,8 +269,7 @@ class MRIFourierCorrected(FourierOperatorBase):
                 r2star_map,
             )
         if self.B is None or self.tl is None:
-            raise ValueError(
-                "Please either provide fieldmap and t or B and tl")
+            raise ValueError("Please either provide fieldmap and t or B and tl")
         self.n_interpolators = self.B.shape[0]
 
         # create spatial interpolator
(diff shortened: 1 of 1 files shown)


src/mrinufft/operators/off_resonance.py
```diff
@@ -219,6 +219,7 @@ def __init__(
         r2star_map=None,
         B=None,
         tl=None,
+        negate=True,
```
chaithyagr, line 222:
```suggestion
        isign=-1,
```

I feel its better to use `isign` which can be `-1` or `1`. Do make sure to check if it is either -1 or 1, else please raise an error

src/mrinufft/operators/off_resonance.py
```diff
@@ -364,11 +368,12 @@ def _get_complex_fieldmap(b0_map, r2star_map=None):
...
 
-def _get_spatial_coefficients(field_map, tl):
+def _get_spatial_coefficients(field_map, tl, negate=True):
     xp = get_array_module(field_map)
-
+    exponent = get_array_module(field_map)
+    exponent = -tl if negate else tl
     # get spatial coeffs
-    C = xp.exp(-tl * field_map[..., None])
+    C = xp.exp(exponent * field_map[..., None])
```
chaithyagr, line 376: You can have it as `isign * tl*field_map`

chaithyagr (changes requested):
Thank you for this. minor comments

chaithyagr (changes requested):

chaithyagr:
Also please make sure to get all the tests to pass. Currently it's just some coding style issue. You can use black or ruff to check and fix it

src/mrinufft/operators/off_resonance.py
Daval-G: @chaithyagr I tend to prefer the original way with a boolean. Here with an integer we should be checking that it's only -1 or 1, and we have to state it explicitly in the docstring while still adding more confusion. The boolean has the advantage of being fail safe, while the integer mostly discharge the developpers of half a line of code to add 3 more and put that charge on the user instead

Daval-G: > Either `-1` (default, e^{-iωt}) or `1` (e^{iωt}).

That line in particular is confusing. Are we speaking about the forward operator or the inverse operator ? Inverse is used to get an image so it would be fair to assume one or the other

chaithyagr: Hmm, I think i agree with you... SOrry for the confusion @femaleprog , can you please revert it?

Issue #248, opened 2025-03-21
MRIFourierCorrected.adj_op should handle shape mismatch between b0_map and smaps internally


femaleprog:
### 🐛 Bug: `MRIFourierCorrected.adj_op` does not handle shape mismatch between `b0_map` and `smaps`

**Description**

When using `MRIFourierCorrected` from `mrinufft.operators.off_resonance`, a shape mismatch error occurs if the B0 map (`b0_map`) and sensitivity maps (`smaps`) have different spatial resolutions.

In typical workflows:
- `smaps` are often computed at high resolution (e.g., `(32, 384, 384, 208)`),
- `b0_map` is typically acquired at lower resolution (e.g., `(64, 64, 32)`).

Currently, this causes a broadcasting error in:
```python
self.C[idx] * self._fourier_op.adj_op(...)

paquiteau:
How do you suggest we address this error? should we upsample the B0 map to match the desired shape, raise warnings /errors earlier on ?

femaleprog:
yes upsampling B0map internally is good, i can make a PR for that later

PR #259, opened 2025-05-05
Fix: Automatically upsample b0_map in MRIFourierCorrected if shapes m…

femaleprog:
## 📝 PR Description

This pull request addresses [Issue #248](https://github.com/mind-inria/mri-nufft/issues/248) — where a shape mismatch between `b0_map` and `smaps` causes a broadcasting error in `MRIFourierCorrected.adj_op`.

### 🔧 What’s fixed

- Automatically upsample `b0_map` using `scipy.ndimage.zoom` when its shape does not match `fourier_op.shape`
- If a `mask` is provided, it is also upsampled accordingly
- A clear `UserWarning` is issued to notify the user that upsampling was applied
- Keeps the operator robust and allows usage without preprocessing outside the module

### 🔬 Why this matters

- In typical MRI workflows:
  - `smaps` are acquired at high resolution (e.g. `(32, 384, 384, 208)`)
  - `b0_map` is often acquired at lower resolution (e.g. `(64, 64, 32)`)
- This mismatch previously caused runtime errors due to shape incompatibility
- Now, the operator handles this internally, simplifying user pipelines and improving reliability

### 🧪 Validation

- Tested on real MRI dataset
- Manual resampling code was removed from the reconstruction pipeline
- Confirmed that the corrected `MRIFourierCorrected` works end-to-end without shape errors

### ⚠️ Runtime warning

```python
UserWarning: B0 map shape (64, 64, 32) does not match operator shape (384, 384, 208).
Upsampling will be performed automatically. Please ensure orientation is correct.

### DIFF
--- a/src/mrinufft/operators/off_resonance.py
+++ b/src/mrinufft/operators/off_resonance.py
@@ -13,6 +13,8 @@ from .._utils import get_array_module
 from .base import FourierOperatorBase
 
 from .interfaces.utils import is_cuda_array
+import warnings
+from scipy.ndimage import zoom
 
 if CUPY_AVAILABLE:
     import cupy as cp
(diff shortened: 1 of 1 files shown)


paquiteau:
looks great ! May be you could add a test case for it as well ?

Issue #258, opened 2025-05-05
Enable batched sensitivity maps in NUFFT operator via n_batchs parameter


AsmaTANABEN:
When applying the NUFFT operator to multi-contrast or batched MRI data (assuming the same undersampling pattern across volumes), I set n_batchs = n_contrasts. In this context, it’s important that the NUFFT operator’s smaps setter supports input of shape (n_batchs, n_coils, *shape), since sensitivity maps vary across contrasts or batch elements.

Currently, I need to instantiate separate NUFFT operators for multi-contrast data and restrict the batch size to 1 for mono contrast dataset, which limits scalability. This feature would make the operator more versatile for multi-contrast reconstruction tasks and for training deep learning models with batch sizes greater than one.

paquiteau:
I see your point, and its a valid one, however this would requires to rewrite the implementation for each backend (!), so not so easy to fix as is. Also note that the current batch implementation is sequential.

However smaps can be updated, so you have an easy workaround by defining something like that:

```python
class BatchedSmapsNufft(FourierOperatorBase):
    ... # init (self.nufft, self.batch_size, etc )
    def op(self.batched_smaps, batched_contrast):
        batched_kspace = self.xp.zero(self.batched_size, self.n_coils, self.nufft.n_samples)
        for b in range(self.batch_size):
            self.nufft.smaps = batch_smaps[b] # or a more lower level copy 
            batched_kspace[b] = self.nufft.op(batch_contrast[b])

       return batched_kspace
```

AsmaTANABEN:
I agree that this isn't a simple fix. Regarding the suggested workaround, I'm a bit unsure how it would interact with backpropagation; specifically, how confident are we that this way of updating smaps  doesn’t interfere with gradient computation? Wouldn’t the operator use the last updated smaps during backprop, potentially leading to incorrect gradients for earlier batch elements? 
Please feel free to correct me if I’m mistaken. I'm bringing this point up because I'm not entirely sure how the current implementation handles this.

paquiteau:
Backprop is likely going to be an issue indeed.  Then I think the `BatchedSmapsNufft` that I proposed above should instead be a `torch.nn.Module` with proper `forward` and `backward` methods. The smaps should be saved (with `ctx.save_for_backward` that is being called in the forward). 
[...]

PR #263, opened 2025-05-19
Add batched smaps support to MRINufftAutograd : Address #258 

AsmaTANABEN:
This PR addresses : [#issue 258](https://github.com/mind-inria/mri-nufft/issues/258), which requests support for batched sensitivity maps (smaps) in the MRINufftAutograd module.

✅ **Fixes / Changes**: 

**-** **Introduced BatchedNufftAutograd**: A new torch.nn.Module extending MRINufftAutograd, allowing the NUFFT operator to handle batched inputs with corresponding batched smaps, while retaining autograd support.
**-** **Updated get_operator function** (mrinufft/operators/base.py): Added 'use_batched_mode' and 'batch_size' as parameters, to dynamically select between the wrappers MRINufftAutograd and the new BatchedNufftAutograd.
**- Extended testing** in test_autodiff.py: Update unit tests to validate the batched mode when use_batched_mode=True.

 **Impact/Use cases :** 

-  **To use the BatchedNufftAutograd wrapper** (when autograd is enabled with respect to data or trajectory), initialize the operator as follows:
`nufft = mrinufft.get_operator(
    backend_name="gpunufft" or "cufinufft", 
    wrt_data=True, 
    wrt_traj=False,
    use_batched_mode=True,
    batch_size=batch_size)(**kwargs)`
 Setting use_batched_mode=True along with batch_size ensures that BatchedNufftAutograd is used instead of the default MRINufftAutograd

**- Enables:**
Calling the NUFFT operator with both batched inputs and batched smaps:
`nufft.adj_op(batched_smaps, batched_kspace_data)
nufft.op(batched_smaps, batched_imgs)`
The output includes an additional batch dimension compared to the original MRINufftAutograd.

**- Supports:**
**Batched training** for deep learning models (e.g., using mono-contrast datasets)
(✅ Tested with my NCPDNet implementation.)
**Single-batch training with multi-contrast data**, where the batch dimension represents multiple contrast channels (e.g., mp2rage imaging data).


paquiteau:
I like this very much ! Two (minors) points of concern however:
- I would rename "use_batched_mode" to something more  explicit like `paired_batch_mode` (we already have batch support)
- It seems that the commit history is a bit messed up (the modification in the io module should not be in this PR)

src/mrinufft/io/siemens.py
paquiteau, line 1: should probably not be in this PR

src/mrinufft/operators/autodiff.py
```diff
@@ -152,3 +152,94 @@ def samples(self, value):
...
+    nufft : object
+        An instance of a standard NUFFT operator.
+    batch_size : int
+        Number of batches to process simultaneously.
+    **kwargs : dict, optional
+        Additional arguments.
+
+    Notes
+    -----
+    #TODO Future improvements may include support for varying trajectories across batches.
```
paquiteau, line 173: Don't put this in the Docstring (A comment is fine though)

Inside the note you can details the math if you wish, and provide an example of usage as well

src/mrinufft/operators/autodiff.py
```diff
@@ -152,3 +152,94 @@ def samples(self, value):
...
+        self._check_input_shape(imgs=batched_smaps)
+        batched_imgs = []
+        for i in range(self.batch_size):
+            try:
+                self.nufft_op.smaps = batched_smaps[i]
+                batched_imgs.append(
+                    _NUFFT_ADJOP.apply(batched_kspace[i], self.samples, self.nufft_op)
+                )
+            except Exception as e:
+                raise RuntimeError(f"Failed at batch index {i+1}: {e}")
```
paquiteau, line 212:
```suggestion
            except Exception as e:
                raise RuntimeError(f"Failed at batch index {i+1}") from e
```

src/mrinufft/operators/base.py
```diff
@@ -286,9 +308,20 @@ def make_autograd(self, wrt_data=True, wrt_traj=False):
         if not self.autograd_available:
             raise ValueError("Backend does not support auto-differentiation.")
 
-        from mrinufft.operators.autodiff import MRINufftAutoGrad
```
paquiteau, line 289: You can do the two imports here

tests/helpers/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .factories import (
     kspace_from_op,
     image_from_op,
+    batchedSmpas_from_op,
```
paquiteau, line 7: typo

tests/helpers/factories.py
```diff
@@ -19,28 +19,42 @@
...
+    kspace = (
+        1j * np.random.randn(*batch_dim, operator.n_coils, operator.n_samples)
+    ).astype(operator.cpx_dtype)
+    kspace += np.random.randn(*batch_dim, operator.n_coils, operator.n_samples).astype(
         operator.cpx_dtype
     )
     return kspace
 
 
+def batchedSmpas_from_op(operator):
```
paquiteau, line 46: typo

tests/operators/test_autodiff.py
paquiteau, line 1: The tests are quite long and complex to read, I would advise to refactor and split them between batched and non-batched version

src/mrinufft/operators/autodiff.py
```diff
@@ -152,3 +152,94 @@ def samples(self, value):
     def __getattr__(self, name):
         """Forward all other attributes to the nufft_op."""
         return getattr(self.nufft_op, name)
+
+
+class BatchedNufftAutoGrad(MRINufftAutoGrad):
```
chaithyagr, line 157: I am curious, do we need a different class for this? Why not support it in the main class itself? Other option is just to add `batch_size` as an argument to make_autograd.

AsmaTANABEN: I believe separating them helps keep each class clean and easier to extend. My initial choice was to keep MRINufftAutoGrad as the base class without adding unnecessary complexity, but I'm open to merging the two classes. @paquiteau what do you think?

paquiteau: After some more though about it, I guess we could merge the two classes together (but keep things tidy with `_{adj}_op_batched` methods if needed) . However. it is not quite clear to me if the smaps are and/or need to be saved in context for this batched version to work

paquiteau: great ! A comment in the code could be added to explicit that :)

src/mrinufft/operators/autodiff.py
```diff
@@ -109,9 +109,17 @@ class MRINufftAutoGrad(torch.nn.Module):
     Parameters
     ----------
     nufft_op: Classic Non differentiable MRI-NUFFT operator.
+
+    batch_size : int, default None
+        Number of batches to process simultaneously.
+        if Provided, the Nufft operator will support different data/smaps pairs
```
chaithyagr, line 115:
```suggestion
        if provided, the NUFFT operator will support different processing
        different k-space / image data and corresponding sensitivity map pairs.
        This is particularly useful for multi-contrast reconstructions
```

src/mrinufft/operators/autodiff.py
```diff
@@ -109,9 +109,17 @@ class MRINufftAutoGrad(torch.nn.Module):
...
+            in a batched mode
+
     """
 
-    def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
+    # TODO Future improvements may include support for varying trajs across batches.
```
chaithyagr, line 120: Please remove this and make it an issue

src/mrinufft/operators/autodiff.py
```diff
@@ -127,15 +135,59 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
             # used for update also.
             self._samples_torch = torch.Tensor(self.nufft_op.samples)
             self._samples_torch.requires_grad = True
+        self.batch_size = batch_size
+        self.paired_batch_mode = batch_size is not None
```
chaithyagr, line 139: You do have `self.batch_size` always with you, do you really need another boolean?

src/mrinufft/operators/autodiff.py
```diff
@@ -152,3 +204,37 @@ def samples(self, value):
     def __getattr__(self, name):
         """Forward all other attributes to the nufft_op."""
         return getattr(self.nufft_op, name)
+
+    def _check_input_shape(self, *, imgs=None, ksps=None):
```
chaithyagr, line 208: This should be `_check_batch_size` right?

src/mrinufft/operators/autodiff.py
```diff
@@ -127,15 +135,59 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
...
+
+    def adj_op_batched(self, batched_kspace, batched_smaps):
+        """Compute the adjoint batched_kspace -> batched_imgs."""
+        # NUFFT op is saved per batch element in ctx to ensure correct backpropagation.
+        self._check_input_shape(ksps=batched_kspace)
+        self._check_input_shape(imgs=batched_smaps)
+        batched_imgs = []
+        for i in range(self.batch_size):
+            try:
+                self.nufft_op.smaps = batched_smaps[i]
```
chaithyagr, line 183: Careful, you could be doing a lot of copies and throttling. But I am okay with it as alternative is just to make many NUFFT ops, which I expect user to make in any case.

src/mrinufft/operators/base.py
```diff
@@ -50,7 +50,12 @@ def list_backends(available_only=False):
 
 
 def get_operator(
-    backend_name: str, wrt_data: bool = False, wrt_traj: bool = False, *args, **kwargs
+    backend_name: str,
+    wrt_data: bool = False,
+    wrt_traj: bool = False,
+    paired_batch_size: int = None,
```
chaithyagr, line 56: If I understand correctly `paired_batch_size` is used only if we want to process for different smaps right? 

Can we rename in that case? `smap_batch_size` ? @paquiteau opinions?

tests/helpers/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .factories import (
     kspace_from_op,
     image_from_op,
+    batchedSmaps_from_op,
```
chaithyagr, line 7:
```suggestion
    batched_smaps_from_op,
```
Lets follow the same naming conventions :)

tests/operators/test_autodiff.py
```diff
@@ -109,30 +170,79 @@ def test_adjoint_and_grad(operator, interface):
...
     )
-
     # Check if nufft and ndft w.r.t trajectory are close in the backprop
     gradient_ndft_ktraj = torch.autograd.grad(
         loss_ndft, operator.samples, retain_graph=True
     )[0]
     gradient_nufft_ktraj = torch.autograd.grad(
         loss_nufft, operator.samples, retain_graph=True
     )[0]
+    # FIXME: atol=5e-1 is too loose?
```
chaithyagr, line 180: Oh yes indeed. Lets discuss how much difference you have

chaithyagr: It could be possible that you have wrong gradients also

chaithyagr (changes requested):
Thank you for this updates. A bunch of comments and major changes from me.

src/mrinufft/operators/base.py
paquiteau: From my understanding in this "paired" mode, the batch size of smaps and data is the same. Therefore I don't think we need an extra integer, a simple boolean flag `paired_batch` (or `paired_batch_mode` fwiw) should be enough ?

src/mrinufft/operators/autodiff.py
```diff
@@ -127,15 +159,118 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
...
+
+    def op(self, x, smaps=None, samples=None):
+        r"""Compute the forward image -> k-space.
+
+        Parameters
+        ----------
+        x: Tensor
+            Image data of shape ``(batch_size, nufft_op.n_batchs, (1 if smaps
+            else nufft_op.n_coils), *nufft_op.shape)``
+        smaps: Tensor, optional
```
chaithyagr: Why tensor? I think this isn't how others are documented right?

src/mrinufft/operators/autodiff.py
```diff
@@ -127,15 +159,118 @@ def __init__(self, nufft_op, wrt_data=True, wrt_traj=False):
...
+            Sensitivity maps of shape ``(batch_size, nufft_op.n_coils, *nufft_op.shape)
+        samples: Tensor, optional
+            Samples for the batches of shape ``(batch_size, nufft_op.n_samples,
+            nufft_op.ndim)``
+
+        Returns
+        -------
+        Tensor:
```
chaithyagr: I see this at many places

src/mrinufft/operators/base.py
```diff
@@ -559,9 +583,20 @@ def __repr__(self):
         )
 
     @classmethod
-    def with_autograd(cls, wrt_data=True, wrt_traj=False, *args, **kwargs):
+    def with_autograd(
+        cls,
+        wrt_data=True,
+        wrt_traj=False,
+        paired_batch_size=None,
```
chaithyagr: Wait here it's banned something else, I'm getting confused

tests/helpers/__init__.py
chaithyagr: I see this still?

chaithyagr (changes requested):
Thanku you for taking it forward. Some old comments still aren't addressed and i found some major changes required in docs. Finally, do we have tests working to test this?

PR #266, opened 2025-05-20
Nifti affine matrix

j-obriot:
This PR adds a nifti_affine function which expands on siemens_quat_to_rot_mat to include scaling and offset.

The affine matrix is usually used by preprocessing and analysis software and it is crucial for registration for example or any step that expect to apply the matrix and get milimeter coordinates.

### issues

The offset in this fuction will usually be wrong by 1/2 of 1 voxel size, from the test I did I couldn't figure out a pattern.

### DIFF
--- a/src/mrinufft/io/siemens.py
+++ b/src/mrinufft/io/siemens.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import numpy as np
-from .utils import siemens_quat_to_rot_mat
+from .utils import siemens_quat_to_rot_mat, nifti_affine
 
 
 def read_siemens_rawdat(
@@ -76,6 +76,7 @@ def read_siemens_rawdat(
         "n_slices": int(twixObj.image.NSli),
         "n_average": int(twixObj.image.NAve),
         "orientation": siemens_quat_to_rot_mat(twixObj.image.slicePos[0][-4:]),
+        "affine": nifti_affine(twixObj),
         "acs": None,
     }
     if "refscan" in twixObj.keys():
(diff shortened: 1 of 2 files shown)


paquiteau:
Hello Joseph, thanks for the fix :)

to determine what the best settings (corner or center of the voxel), maybe you could compare with the affine provided in DICOM images generated from the scanner ? that way we have the same convention in k-space files and images files.

j-obriot:
Hello,

Yes that's what I did but from the (very little) testing I did, the offset of 0, 1/2 or 1 seemed to vary between configurations and dimensions. I will be comparing with various data files soon to try to elucidate that.

chaithyagr:
I had a discussion with @j-obriot , sadly the weird 1 or 1/2 slice shift offset is kinda random at least for now. But I feel even if we don't understand it, this PR is surely something better than nothing (what we have currently). We could open a different issue to track this weird extra shift

j-obriot:
Just a quick update,

Here is a comparison of the origin point from different reconstruction from the scanner vs their raw file's ggrappa reco and extracted matrices using this PR.

1. 3D EPI
2. 3D GRE (partition axial)
3. 3D GRE (partition sagital)

In particular, there is an issue with the original siemens_quat_to_rot_mat function giving a wrong orientation matrix (or maybe the issue lies within ggrappa reconstruction?). At least the EPI reco is shifted one slice down compared to the scanner reco. 

affine

chaithyagr:
I think we can open another issue to solve other parts of this problem as I expect it wont be straightforward

chaithyagr:
I just re-ran the tests. It was failing due to finufft upgrades which got cached in your runs. I will handle these failures in #224

Issue #273, opened 2025-06-05
Add support for handeling slew issues for UTE trajectories


chaithyagr:
In #269 we updated the trajectory spec, allowing for having a buffer before switching on ADC. This allows to better handle pre-phsors and rewinders.
However, the UTE trajectories were not fully handled.
In particular, things that could be addressed:

- [ ] https://github.com/mind-inria/mri-nufft/pull/269#issuecomment-2934306783
- [ ] https://github.com/mind-inria/mri-nufft/pull/269#discussion_r2128064542

PR #276, opened 2025-06-19
Add prephasors, rewinders etc.

chaithyagr:
This is just the same PR #269 . Adding back @Daval-G and @j-obriot for review.


src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +378,238 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+
+    return n_ramp_down, n_ramp_up, n_plateau, gi
+
+
+def get_gradient_amplitudes_to_travel_for_set_time(
+    ke: NDArray,
+    ks: NDArray | None = None,
+    gs: NDArray | None = None,
+    ge: NDArray | None = None,
+    N: int | None = None,
```
Daval-G, line 480: That name is too simple for a public function

Daval-G (changes requested):
As usual great work, just a bunch of small points to discuss but there should be few iterations

src/mrinufft/io/nsp.py
```diff
@@ -226,10 +243,24 @@ def write_trajectory(
...
+        The ratio of trajectory when TE occurs, with 0 as start of
+        trajectory and 1 as end. By default 0.5, which is the
+        center of the trajectory (in-out trajectory).
     gmax : float, optional
         Maximum gradient magnitude in T/m, by default 0.04
     smax : float, optional
         Maximum slew rate in T/m/ms, by default 0.1
+    pregrad : str, optional
+        Pregrad method, by default `prephase`
+        `prephase` will add a prephase gradient to the start of the trajectory.
```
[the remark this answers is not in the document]
Daval-G: Oh and it's prephase instead of pregrad in the description. Why that change ?

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +378,238 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+    raster_time : float, optional
+        Time interval between gradient samples (s). Default is DEFAULT_RASTER_TIME.
+    gmax : float, optional
+        Maximum gradient amplitude (T/m). Default is DEFAULT_GMAX.
+    smax : float, optional
+        Maximum slew rate (T/m/s). Default is DEFAULT_SMAX.
+
+    Returns
+    -------
+    G : NDArray
```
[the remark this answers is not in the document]
chaithyagr: CoPiLoT xD

src/mrinufft/trajectories/tools.py
```diff
@@ -420,23 +420,34 @@ def get_gradient_times_to_travel(
...
     Returns
     -------
-    n_ramp_down: The timing values for the ramp down phase.
-    n_ramp_up: The timing values for the ramp up phase.
-    n_plateau: The timing values for the plateau phase.
-    gi: The intermediate gradient values for trapezoidal or triangular waveforms.
+    The timing values for the ramp down phase.
+    The timing values for the ramp up phase.
+    The timing values for the plateau phase.
+    The intermediate gradient values for trapezoidal or triangular waveforms.
```
paquiteau: you can still names the return values, it helps understanding the code directly 

https://numpydoc.readthedocs.io/en/latest/format.html#returns

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +378,353 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+    raster_time: float = DEFAULT_RASTER_TIME,
+    gmax: float = DEFAULT_GMAX,
+    smax: float = DEFAULT_SMAX,
+) -> tuple[NDArray, NDArray, NDArray, NDArray]:
+    """Get gradient timing values for trapezoidal or triangular waveforms.
+
+    Compute gradient timing values to take k-space trajectories
+    from position ``ks`` with gradient ``gs`` to position ``ke`` with gradient
+    ``ge``, while being hardware compliant.
+    This function calculates the number of time steps required for the ramp down,
```
paquiteau, line 396:
```suggestion
    This function calculates the minimal number of time steps required for the ramp down,
```

src/mrinufft/trajectories/tools.py
paquiteau, line 1: I think these two functions would benefit from longer in-code comments, taking the reader by the hand (First, we are going to compute this, then that because this , etc...). It does not have to be in the docstring though.

paquiteau (changes requested):
Here are some general comments, I don't understand it fully, but I do see that is great stuff nonetheless !

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +378,353 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+        2 * area_needed[direct_possible_mask]
+        - (direct_ramp_down + 1) * start_gradients[direct_possible_mask]
+        - (direct_ramp_up - 1) * end_gradients[direct_possible_mask]
+    ) / (direct_ramp_down + direct_ramp_up)
+    gi[direct_possible_mask] = gi_view
+    n_plateau[direct_possible_mask] = 0
+    n_ramp_down[direct_possible_mask] = direct_ramp_down
+    n_ramp_up[direct_possible_mask] = direct_ramp_up
+    # Replace NaNs in any calculation above
+    gi[np.isnan(gi)] = start_gradients[np.isnan(gi)]
```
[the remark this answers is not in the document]
chaithyagr: Nope, if it is NaN it means end_gradients and start_gradients are the same. So gi is also start gradients. But i guess linear interpolation also will be the same.

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+
+    gi = Parallel(n_jobs=n_jobs)(
+        delayed(solve_gi_min_plateau)(
+            start_gradients[i, j],
+            end_gradients[i, j],
+            area_needed[i, j],
+        )
+        for i in range(start_gradients.shape[0])
+        for j in range(start_gradients.shape[1])
+    )
```
paquiteau, line 494:
```suggestion
    gi = Parallel(n_jobs=n_jobs)(
        delayed(solve_gi_min_plateau)(gs,ge,area) 
        for gs,ge,area in zip(start_gradients[:],end_gradients[:], area_needed[:]))
```
more pythonic :) I am almost sure that a full vectorized method is also possible, but I don't deem this to be too critical. 

The main question is if this takes some non-negligeable times or not ? (e.g. is joblib really necessary here)

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+def get_gradient_times_to_travel(
+    kspace_end_loc: Optional[NDArray] = None,
+    kspace_start_loc: Optional[NDArray] = None,
+    end_gradients: Optional[NDArray] = None,
+    start_gradients: Optional[NDArray] = None,
+    gamma: float = Gammas.Hydrogen,
+    raster_time: float = DEFAULT_RASTER_TIME,
+    gmax: float = DEFAULT_GMAX,
+    smax: float = DEFAULT_SMAX,
+    n_jobs: int = 1,
```
paquiteau, line 426:
```suggestion
    n_jobs: int = -1,
```
As per https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+    """
+    area_needed = (kspace_end_loc - kspace_start_loc) / gamma / raster_time
+
+    def solve_gi_min_plateau(gs, ge, area):
+        def _residual(gi):
+            n_down, n_up = _calculate_ramps(gs, ge, gi, smax, raster_time)
+            n_pl = _calculate_plateau(gs, ge, gi, n_down, n_up, area)
+            if n_pl < 0:
+                return np.abs(n_pl) * 10000  # Penalize negative plateau
+            return n_pl * 100
```
paquiteau, line 475: Why the 100 factor ? 
Also why are negative plateau a bad idea ?

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+        Maximum slew rate (T/m/s). Default is DEFAULT_SMAX.
+    n_jobs : int, optional
+        Number of parallel jobs to run for optimization, by default 1.
+
+    Returns
+    -------
+    NDArray
+        Gradient waveforms, shape (nb_shots, nb_samples_per_shot, nb_dimension)
+        , where each entry contains the gradient value at each time step
+         for each shot and dimension.
```
paquiteau, line 568:
```suggestion
        Gradient waveforms, shape (nb_shots, nb_samples_per_shot, nb_dimension),
        where each entry contains the gradient value at each time step for each shot
        and dimension.
```

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+    n_jobs : int, optional
+        Number of parallel jobs to run for optimization, by default 1.
+
+    Returns
+    -------
+    n_ramp_down: The timing values for the ramp down phase.
+    n_ramp_up: The timing values for the ramp up phase.
+    n_plateau: The timing values for the plateau phase.
+    gi: The intermediate gradient values for trapezoidal or triangular waveforms.
+    """
```
paquiteau, line 466:
```suggestion
    See Also
    --------
    get_gradient_amplitudes_to_travel_for_set_time :
        To directly get the waveforms required. This is most-likely what you want to use.
    """
```

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+
+
+def _calculate_ramps(gs, ge, gi, smax, raster_time, ceil=False, buffer=0):
+    """Calculate the number of time steps for the ramp down and up."""
+    n_ramp_down = np.abs(gi - gs) / (smax * raster_time)
+    n_ramp_up = np.abs(ge - gi) / (smax * raster_time)
+    if ceil:
+        n_ramp_down = np.ceil(n_ramp_down).astype(int)
+        n_ramp_up = np.ceil(n_ramp_up).astype(int)
+    return n_ramp_down + buffer, n_ramp_up + buffer
```
paquiteau, line 407: why the buffer ? Also in the rest of the code it is always equal to 1, consider just inlining it ? (with a comment to say why this buffer is required)

src/mrinufft/trajectories/tools.py
```diff
@@ -369,6 +380,290 @@ def unepify(trajectory: NDArray, Ns_readouts: int, Ns_transitions: int) -> NDArr
...
+                start_gradients[i, d], gi[i, d], n_ramp_down[i, d], endpoint=False
+            )
+            start += n_ramp_down[i, d]
+            if n_plateau[i, d] > 0:
+                G[i, start : start + n_plateau[i, d], d] = gi[i, d]
+                start += n_plateau[i, d]
+            G[i, start : start + n_ramp_up[i, d], d] = np.linspace(
+                gi[i, d], end_gradients[i, d], n_ramp_up[i, d], endpoint=False
+            )
+    return G
```
paquiteau, line 664: linspace is vector-friendly:

```suggestion
    G = np.zeros((nb_shots, nb_raster_points, nb_dimension), dtype=np.float32)
    for i in range(nb_shots):
        start = n_ramp_down[i,0]
        G[i, : start] = np.linspace(
            start_gradients[i], gi[i], n_ramp_down[i], endpoint=False, axis=-1
        )
        if n_plateau[i, d] > 0:
            G[i, start : start + n_plateau[i, 0]] = gi[i]
            start += n_plateau[i, 0]
        G[i, start : start + n_ramp_up[i, 0]] = np.linspace(
            gi[i], end_gradients[i], n_ramp_up[i, 0], axis=-1, endpoint=False
        )
    return G
```

tests/test_io.py
```diff
@@ -17,68 +25,146 @@ def case_trajectory_2D(self):
...
+            0,
             5,
-            10e3,
+            Gammas.Na,
             1.2,
         )
 
 
+@parametrize("gamma", [Gammas.Hydrogen])
+@parametrize("raster_time", [DEFAULT_RASTER_TIME])
```
paquiteau, line 48:
```suggestion
@parametrize("gamma,raster_time", [(Gammas.Hydrogen, DEFAULT_RASTER_TIME)])
```

tests/test_io.py
```diff
@@ -17,68 +25,146 @@ def case_trajectory_2D(self):
...
+    assert np.all(np.abs(GE[:, -1]) / raster_time < smax)
+    assert GE.shape[1] < 200  # Checks to ensure we don't have too many samples
+    # Check that ending location matches.
+    np.testing.assert_allclose(
+        0,
+        trajectory[:, -1] + np.sum(GE, axis=1) * gamma * raster_time,
+        atol=1e-2 / min(resolution) / 2,
+    )
+    # Check that gradients match.
+    np.testing.assert_allclose(GE[:, 0], gradients[:, -1], atol=1e-5)
```
paquiteau, line 111: Consider putting this in a dedicated test to have smaller tests ?

src/mrinufft/io/nsp.py
```diff
@@ -261,6 +333,15 @@ def write_trajectory(
                 f"Maximum gradient amplitude: {maxG:.3f} > {gmax:.3f}"
                 f"Maximum slew rate: {maxS:.3f} > {smax:.3f}"
             )
+        if pregrad != "prephase":
```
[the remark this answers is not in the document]
chaithyagr: naahh, i disagree.. As @Daval-G mentioned, the sequence can handle pre-phasors right. This warning shouyld come up only in the case of UTE trajectories.. I think I am handeling that in #273

tests/test_io.py
chaithyagr: I dont think i got you. Thing is the test takes time to run as it runs solver internally. I would prefer having all the checks in same test due to this. Lemme know if you know how to refactor to pre-calculate and then do checks in smaller tests

src/mrinufft/trajectories/tools.py
chaithyagr: Ohh no no, I want it to run one single thread by default.. its slower for most cases to use multi-threading unless you are working on large sampling patterns.

src/mrinufft/trajectories/tools.py
chaithyagr: It is not negative plateau, but rather negative discrete time events during which plateau occurs. By definition it should be greater than 0.
In fact all `n_*` needs to be >0 and integers. (n_ramp_up, n_ramp_down, n_pl)

src/mrinufft/trajectories/tools.py
chaithyagr: I dont understand this, the ramps and gi values can change for each dimension as well. Sadly there is no way to vectorize it. However, this is still pretty fast.

tests/test_io.py
paquiteau: you can use a fixture for the precomputation :)

Daval-G (approved):
I trust @paquiteau's approval and @chaithyagr on recent changes, I will take too much time to review otherwise

PR #283, opened 2025-07-03
Pulseq

### LINKED ISSUE #102: Pulseq Integration ?
At ISMRM 2024, peoples asks if there were any Pulseq integration. 

I am not familiar with Pulseq, but let's have this here for future reference and discussion.

### LINKED ISSUE #262: Add support for pulseq sequences
Currently the only missing component is to have kspace trajectories from mrinufft to run on scanner. For this it would help a lot to have a pulseq sequences generation as an extras module with dependency on pypulseq

This is more of discussion thread for now before we work on it

### LINKED ISSUE #270: Add example showcasing the use of `pregrad` and `postgrad`
In #269 we added new features into mri-nufft: `pregrad` and `postgrad`.
We need examples showcasing its use through gradient plots and possibly trajectory curves.
We could potentially use different colours to showcase `prephasor` and `rewinder`

### LINKED ISSUE #286: Faster prephasors with linear programming
## The problem
In #276 and #276 we added support for designing gradient waveform to connect two point in k-spac effiently, or so I though. While working on new trajectories, I found some MATLAB implementation use to stitches spokes together, and that was using Linear Programming. 

TLDR: I decided to try it and compare with the current implementation: 
- Blue is the trajectory to connect, red is using Linear Programming, green is current method using `get_gradient_amplitudes_to_travel_for_set_time`
- The Linear Programming solution finds shorter path (22 points vs >500) between spokes
- Computing the connecting path is slower (but can be parallellize between spokes)
- Unlike the existing implementation, this does not enforce the use of trapezes for the gradient waveforms. 
Image

A prototype and more insights can be found in: [lip.ipynb](https://github.com/user-attachments/files/21942341/lip.ipynb)

### Math Formulation of Linear Programming

* **Variables**: The gradient values at each time point for the `x`, `y`, and `z` axes.
* **LP Constraints**:
    * **Integral**: The sum of gradient values must equal a target k-space displacement ($\Delta k$).
    * **Slew Rate**: The change between consecutive points must not exceed a maximum rate ($s_{max}$).
    * **Amplitude**: The values must be within a maximum amplitude ($g_{max}$).
    * **Boundary Conditions**: The start and end values are fixed.

All constraints are linear, allowing the use of a standard LP solver. The constraints are formalized as matrices and vectors: see the [Scipy doc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html) for how this is done.

 **However, A big drawback of Linear Programming is that the length of the solution vector must be known**, and if the prescribe size is to small, there is no solution available.

To circumvent that, I implemented a binary search between a lower and upper bound on the number of points required to create the connection, and the goal is to find the smallest solution that is acceptable. 

### More Gotcha's: 

- The method is slow (4s for 256 spokes, vs 0.18 with the existing solution), but can be parallelize (e.g. using joblib) to perform multiple connections.
- differents connections path may have different length (the current implementation enforce the same lenght for all connection to compute), which could create some problems for echo spacing for a trajectory that passes multiple time in the center during a single (connected) shot.
- Getting the number of points right is tricky, if it fails outside the currently provided bounds, it fails hard. Getting tighter bounds also help the binary search to go faster. 
- I think some edge cases (cartesian displacement for instance) could be done indepently, and faster (instead of solving of gx,gy,gz, we only have to solve the simpler problem for gx instead, the other waveform being zeroed.)
- LLM were use in the design of this, and I am not an expert in Linear Programming.

## Next steps
- I would like some feedback on this methodology, and help on checking that it does work as we expect.  
- Open a PR with 
- Write some more lengthy documentation (that could take the form of a blog post) 

## References

- Linear Programming: https://en.m.wikipedia.org/wiki/Linear_programming
- Usage in MRI: I think I have seen a paper talking about it, but can't find it again....

paquiteau:
- Closes #102
- Closes #262 
- Closes #286
- Closes #270

- [x] Add a reader for pulseq sequences to extract kspace trajectories (formatted to MRI-nufft conventions) from pulseq files
   - That would make for a nice functional test `traj -> write_seq -> read_seq -> compare`.
- ~~( Add a helper function to create gradients block for pypulseq)~~
   -  there is now a `prepare_trajectory_for_seq` that creates gradients waveforms with pre/re-winders. Generating the gradients event directly does not make much sense I feel. 
- [x] Add a simple generation of a GRE sequence with pypulseq.

### DIFF
--- a/src/mrinufft/io/pulseq.py
+++ b/src/mrinufft/io/pulseq.py
@@ -25,6 +25,7 @@ try:
     import pypulseq as pp
 except ImportError:
     PULSEQ_AVAILABLE = False
+    pp = None
 
 
 def read_pulseq_traj(
@@ -116,7 +117,7 @@ def _check_timings(seq):
         warnings.warn("Timing check failed. Error listing follows:" + str(error_report))
 
 
-def acq2opts(acq: Acquisition) -> pp.Opts:
+def acq2opts(acq: Acquisition) -> "pp.Opts":
     """Convert an Acquisition object to pypulseq Opts.
 
     Parameters
--- a/src/mrinufft/io/pulseq.py
+++ b/src/mrinufft/io/pulseq.py
@@ -1,0 +1,326 @@
+"""Pulseq Trajectory Reader and writers.
+
+Reads Pulseq `.seq` files to extract k-space trajectory and other parameters. It
+also provides functionality to create pulseq block and shape objects to
+facilitate the integration of arbitrary k-space trajectories into Pulseq
+sequences. Requires the `pypulseq` package to be installed.
+"""
+
+import warnings
+from types import SimpleNamespace
+
+import numpy as np
+from numpy.typing import NDArray
+
+from mrinufft.trajectories.utils import (
+    Acquisition,
+    Hardware,
+    convert_trajectory_to_gradients,
+)
+
+from mrinufft.trajectories.gradients import get_prephasors_and_spoilers
+
+PULSEQ_AVAILABLE = True
+try:
+    import pypulseq as pp
+except ImportError:
+    PULSEQ_AVAILABLE = False
+
+
+def read_pulseq_traj(
+    seq, trajectory_delay=0.0, gradient_offset=0.0
+) -> tuple[NDArray, dict, Acquisition]:
+    """Extract k-space trajectory from a Pulseq sequence file.
+
+    The sequence should be a valid Pulseq `.seq` file, with arbitrary gradient
+    waveforms, which all have the same length.
+
+
+    Parameters
+    ----------
+    sequence : pulseq Sequence, or Path to the Pulseq `.seq` file.
+        The Pulseq sequence object or the path to the `.seq` file.
+    trajectory_delay : float, optional
+        Compensation factor in seconds (s) to align ADC and gradients.
+    gradient_offset : float, optional
+        Simulates background gradients (specified in Hz/m)
+
+    Returns
+    -------
+    NDArray
+        The k-space trajectory as a numpy array of shape (n_shots, n_samples, 3),
+        where the last dimension corresponds to the x, y, and z coordinates in k-space.
+    dict
+        the [DESCRIPTION] block from the sequence file.
+    Acquisition
+        The acquisition parameters extracted from the sequence.
+
+    """
+    if not PULSEQ_AVAILABLE:
+        raise ImportError(
+            "The `pypulseq` package is required for this function. "
+            "Please install it via `pip install pypulseq` or "
+            "`pip install mri-nufft[io]`."
+        )
+    if not isinstance(seq, pp.Sequence):
+        filename = seq
+        seq = pp.Sequence()
+        seq.read(filename)
+
+    kspace_adc, _, t_exc, t_refocus, t_adc = seq.calculate_kspace(
+        trajectory_delay=trajectory_delay, gradient_offset=gradient_offset
+    )
+
+    if not len(t_adc):
+        raise ValueError(
+            "The sequence does not contain any ADC events. "
+            "Please ensure that the sequence has ADC events defined."
+        )
+    # split t_adc with t_exc and t_refocus, the index are then used to split kspace_adc
+    t_splits = np.sort(np.concatenate([t_exc, t_refocus]))
+    idx_prev = 0
+    kspace_shots = []
+    for t in t_splits:
+        idx_next = np.searchsorted(t_adc, t, side="left")
+        if idx_next == idx_prev:
+            continue
+        kspace_shots.append(kspace_adc[:, idx_prev:idx_next].T)
+        if idx_next == kspace_adc.shape[1] and t > t_adc[-1]:  # last useful point
+            break
+        idx_prev = idx_next
+    if idx_next < kspace_adc.shape[1]:
+        kspace_shots.append(kspace_adc[:, idx_next:].T)  # add remaining gradients.
+    # convert to KMAX standard.
+    #
+    acq = Acquisition(
+        fov=seq.get_definition("FOV"),
+        img_size=seq.get_definition("ImgSize"),
+        adc_dwell_time=seq.system.adc_raster_time,
+        hardware=Hardware(
+            grad_raster_time=seq.system.grad_raster_time,
+            gmax=seq.system.max_grad / 1e3,  # FIXME: assumes mT/m in pulseq
+            smax=seq.system.max_slew,
+        ),
+        gamma=seq.system.gamma / (2 * np.pi),
+    )
+    kspace_shots = np.ascontiguousarray(kspace_shots)
+    return kspace_shots, seq.definitions, acq
+
+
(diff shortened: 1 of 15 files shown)


src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,127 @@
...
+    list
+        A list of Pulseq gradient blocks corresponding to the k-space trajectory.
+    """
+    grads = convert_trajectory_to_gradients(trajectory)
+
+    # TODO add prewind/postwind stuff from #276
+    # TODO ensure amplitudes are correct
+
+    # TODO deduplicate gradients (reduces the size of the file, speed up the sequence loading).
+    # TODO Are rotation of trajectories supported in pulseq?
```
chaithyagr, line 84: yes :) https://github.com/imr-framework/pypulseq/blob/master/src/pypulseq/rotate.py
But i wonder why you need it. The rotation of trajectory to match the GOV rotation is done by the sequence at scanner.

src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,127 @@
...
+    -------
+    list
+        A list of Pulseq gradient blocks corresponding to the k-space trajectory.
+    """
+    grads = convert_trajectory_to_gradients(trajectory)
+
+    # TODO add prewind/postwind stuff from #276
+    # TODO ensure amplitudes are correct
+
+    # TODO deduplicate gradients (reduces the size of the file, speed up the sequence loading).
```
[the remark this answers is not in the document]
paquiteau: yes that was the idea behind this see for more detail: https://github.com/imr-framework/pypulseq/discussions/184

examples/example_pulseq.py
```diff
@@ -0,0 +1,98 @@
...
+"""
+import numpy as np
+from mrinufft.io import pulseq_gre_3D, prepare_trajectory_for_seq
+from mrinufft.trajectories.display import display_3D_trajectory
+from mrinufft.trajectories import initialize_2D_spiral, stack
+
+import matplotlib.pyplot as plt
+
+# %%
+# Some Constant values
```
chaithyagr, line 17:
```suggestion
# %%
# Defining the sequence parameters like repetition time (TR), echo time (TE), flip angle (FA), field-of-view (FOV) and image matrix size.
```

tests/test_io.py
```diff
@@ -178,3 +184,21 @@ def test_add_shift(kspace_loc, shape):
...
+    cases=[CasesTrajectories.case_random3D],
+)
+def test_pulseq_write_read(kspace_loc, FOV, img_size):
+    """Test the Pulseq write and read functions."""
+
+    kspace_loc = kspace_loc.astype(np.float32)
+
+    seq = pulseq_gre_3D(
+        trajectory=kspace_loc, fov=FOV, img_size=img_size, TR=1000, TE=100, FA=90
+    )
```
chaithyagr, line 200: Maybe you can also add checks to see if TE, TR and FA are satisfied?

src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,255 @@
...
+    return seq.definitions, kspace_shots
+
+
+def _check_timings(seq):
+    # Check whether the timing of the sequence is correct
+    ok, error_report = seq.check_timing()
+    if ok:
+        return None
+    else:
+        warnings.warn("Timing check failed. Error listing follows:" + str(error_report))
```
chaithyagr, line 77: Why a separate function really?

src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def pulseq_gre_3D(
+    trajectory: NDArray,
```
chaithyagr, line 81: Why not have a generic one for both 2D and 3D?

src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,255 @@
...
+        - rf_pulse.delay
+        - TE_pos * shot_duration
+        - skip_start * system.grad_raster_time
+    )
+    delay_end_TR = pp.make_delay(
+        TR
+        - pp.calc_duration(rf_pulse)
+        - delay_before_grad.delay
+        - full_grads.shape[1] * system.grad_raster_time
+    )
```
chaithyagr, line 204: You might need to check for min_TE and min_TR. Can you add a util function to get it?
Basically for a fixed trajectory and spoil levels, give the minimum TE and TR possible.

src/mrinufft/io/pulseq.py
```diff
@@ -0,0 +1,255 @@
...
+    adc = pp.make_adc(
+        num_samples=traj_length * osf,
+        system=system,
+        duration=shot_duration,
+        delay=skip_start * system.grad_raster_time,
+    )
+
+    # Add a spoiler gradient that move to twice the edge of k-space in the x direction.
+    spoiler = pp.make_trapezoid(
+        channel="x", area=2 * img_size[0] / fov[0], system=system
```
chaithyagr, line 189: The spoiler needs to be user chooseable. Ideally have this factor `2` as specifiable by user.

chaithyagr (changes requested):
Great work. Some major questions and minor reviews. I will need another round of review.

src/mrinufft/io/pulseq.py
paquiteau: Cause I am planning to add other sequences (like a 2D GRE for instance) this would be common to all of them.

src/mrinufft/io/pulseq.py
paquiteau: 2D GRE have some specificities (specific pulse and slice selection that changes over time), it makes more sense to me to have it separately (some core  will be refactor if needed)

src/mrinufft/io/pulseq.py
paquiteau: this is handled by the `check_timing` function of pulseq (there is check for delay that can't be negative for instance)

src/mrinufft/io/pulseq.py
paquiteau: okay will do

paquiteau:
Alright I rebased this on top of #297 and master, so this is good for review I think. 

I only provide a 3D GRE example for now, we might propose more sequence buildings block in the future if there is any interest

chaithyagr (changes requested):
Awesome work! I didnt delve a lot into the math, I think the example will give me the confidence and I will fix any issues I see in my UTE PR.

Some comments and question

/build-docs

src/mrinufft/trajectories/gradients.py
```diff
@@ -119,3 +137,662 @@ def _default_update_parameters(shot: NDArray, *parameters: list) -> list:
...
+    N: int | None = None,
+) -> NDArray:
+    """
+    Get the gradient connections for a set of start and end points.
+
+    Parameters
+    ----------
+    $params_connect
+
+    N: int, optional
```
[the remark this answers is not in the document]
paquiteau: nah, i like it as it is ^^

PR #284, opened 2025-08-20
fix: operator shape and try to use gpu.

### LINKED ISSUE #222: Make Conjugate Gradient Descent work on GPU
Currently the `cg` method in `extras/gradient.py` has a with_numpy decoration, meaning that the data will do roundtrip to the CPU. It should be able to run directly on GPU. 

Plan of action:
- Use `with_numpy_cupy` instead
- adapt the computation inside to use the correct array module.

### LINKED ISSUE #253: Not the right operator shape for initialization
We get an issue if someone does cg without SENSE and also if x_init isnt specified.
The problem is that operrator.shape doesnt reflect exactly what is expected as input.
It would help to update `operator.shape` in base.py

https://github.com/mind-inria/mri-nufft/blob/cbcdf2abfb61f97047d15f6de3d67af3c4e9708f/src/mrinufft/extras/gradient.py#L39-L43

paquiteau:
- Closes #253
- Closes #222 

- use with_numpy_cupy for wrapping CG
- correct shape initialization if x_init is not provided.

### DIFF
--- a/src/mrinufft/extras/gradient.py
+++ b/src/mrinufft/extras/gradient.py
@@ -68,6 +68,7 @@ def cg(operator, kspace_data, x_init=None, num_iter=10, tol=1e-4, compute_loss=F
         velocity = grad_new + beta * velocity
 
         image = image - velocity / lipschitz_cst
+        grad = grad_new
         if compute_loss:
             loss.append(calculate_loss(image))
     return image if loss is None else (image, xp.array(loss))
--- a/src/mrinufft/extras/gradient.py
+++ b/src/mrinufft/extras/gradient.py
@@ -1,11 +1,10 @@
 """Conjugate gradient optimization algorithm for image reconstruction."""
 
-import numpy as np
+from mrinufft._array_compat import with_numpy_cupy
+from mrinufft._utils import get_array_module
 
-from mrinufft.operators.base import with_numpy
 
-
-@with_numpy
+@with_numpy_cupy
 def cg(operator, kspace_data, x_init=None, num_iter=10, tol=1e-4, compute_loss=False):
     """
     Perform conjugate gradient (CG) optimization for image reconstruction.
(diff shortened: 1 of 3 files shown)


src/mrinufft/extras/gradient.py
```diff
@@ -36,34 +35,39 @@ def cg(operator, kspace_data, x_init=None, num_iter=10, tol=1e-4, compute_loss=F
...
+    if operator.uses_sense:
+        init_shape = operator.shape
+    else:
+        init_shape = (operator.n_coils, *operator.shape)
     image = (
-        np.zeros(operator.shape, dtype=type(kspace_data[0]))
+        xp.zeros(init_shape, dtype=type(kspace_data[0]))
         if x_init is None
-        else x_init
+        else x_init.reshape(init_shape)
```
alineyyy, line 46: Why do we need reshape here?  I didn't get the point

src/mrinufft/extras/gradient.py
```diff
@@ -36,34 +35,39 @@ def cg(operator, kspace_data, x_init=None, num_iter=10, tol=1e-4, compute_loss=F
...
+        if xp.linalg.norm(grad_new) <= tol:
             break
 
-        beta = np.dot(
+        beta = xp.dot(
             grad_new.flatten(), (grad_new.flatten() - grad.flatten())
-        ) / np.dot(grad.flatten(), grad.flatten())
+        ) / xp.dot(grad.flatten(), grad.flatten())
         beta = max(0, beta)  # Polak-Ribiere formula is used to compute the beta
         velocity = grad_new + beta * velocity
```
alineyyy, line 69: Wait... I found that the update of grad is missed here.
`grad  = grad_new ` should be added after calculating beta

alineyyy (commented):
Thank you for this essential change! It will make CG much faster than before.
Just one thing more: could you please add the step of gradient update inside CG algorithm?  
Everything else looks good to me except this :+1:

src/mrinufft/extras/gradient.py
chaithyagr: Good catch

Issue #272, opened 2025-06-05
Add support for projection step for k-space trajectories based on GMax, SMax values.


chaithyagr:
This is a key missing ingridient in our mri-nufft package. We would like to implement: https://pubmed.ncbi.nlm.nih.gov/27019479/

Daval-G:
I have a working implementation based on some old matlab repository, with fewer features to simplify (for example no need for rotation-variant version when we can apply the rotation-invariant version over each channel individually). Let's sync on that to avoid redundant development efforts

PR #285, opened 2025-08-20
Acquisition Configuration

### LINKED ISSUE #257: Acquisition parameters for trajectories
There are many trajectory designs I would like to add that depend on acquisition parameters (analytic like TWIRL/WHIRL, but also most heuristic and optimization-based methods). However:
(1) Most of them need tons of parameters (max gradients, max slew rates, resolution, field-of-view, sampling rate, etc) on top of the ones really needed for customization. I have some prototype trajectories with 20+ arguments, it's too much.
(2) To keep all trajectories homogeneous, we would have all of them either normalized between -0.5 and 0.5, or to the provided resolutions. The first solution is weird as the generated trajectories would be relevant only to specific acquisition conditions, but the second solution would mean adding an extra parameter to all trajectories just for scaling.

Today our trajectories are all scaled to a global constant KMAX. I was wondering if it would be relevant to let the users change it, similarly to the DisplayConfig ? In addition we could set most acquisition parameters under some AcquisitionConfig. Otherwise we could just have an AcquisitionParameters class to carry. There are tons of possible solutions, any opinion on that ? @chaithyagr @paquiteau

### LINKED ISSUE #282: Create a new dataclass and an enum for scanner parameters
## Description & use cases

We happen to use a lot of parameters defined in utils.py of trajectory: GMAX, SMAX, RASTER_TIME etc. All this should potentially be moved to a dataclass and enum to avoid passing a lot of parameters to all the functions. This can also help in setting a single class for projection trajectories in future #221 .

paquiteau:
This PR introduces an Acquisition Configuration class, to help manage physical units relevant to MR system.

- Closes #282
- Closes #257
- Open a clean path to #272, and all hardware dependent trajectories.
- There are some conflict with #283 , I will take care of it once we know which PR goes first.

- Introduce the Acquisition class
- Update most of the functions (notable exception: `read_trajectory` in `io.nsp` module) that were relying on parameters such as `gmax`, `smax`, `raster_time` to use the acquisition object instead. 

This is a significant change (a lot of function signatures are modified), I would recommend a new release (2.0.0, including pulseq support). 
Note also that I tried to express all physical quantities in raw SI units (a.k.a. no kHz/G/cm)

TODO:

- [ ] Update test to use Acquisition as well 
- [ ] Add a doc file about the Acquisition 
- [ ] Add hardware configuration for well-known scanners (contribution welcome !)

### DIFF
--- a/src/mrinufft/io/nsp.py
+++ b/src/mrinufft/io/nsp.py
@@ -1,3 +1,4 @@
+# type: ignore
 """Read/Write trajectory for Neurospin sequences."""
 
 from __future__ import annotations
@@ -8,13 +9,9 @@ from array import array
 from datetime import datetime
 
 import numpy as np
-from typing import Optional
 
 from mrinufft.trajectories.utils import (
-    DEFAULT_GMAX,
-    DEFAULT_RASTER_TIME,
-    DEFAULT_SMAX,
-    KMAX,
+    Acquisition,
     Gammas,
     check_hardware_constraints,
     convert_gradients_to_slew_rates,
@@ -207,19 +204,13 @@ def _pop_elements(array, num_elements=1, type=np.float32):
 
 def write_trajectory(
     trajectory: np.ndarray,
-    FOV: tuple[float, ...],
-    img_size: tuple[int, ...],
     grad_filename: str,
-    norm_factor: float = KMAX,
-    gamma: float = Gammas.HYDROGEN,
-    raster_time: float = DEFAULT_RASTER_TIME,
     check_constraints: bool = True,
     TE_pos: float = 0.5,
-    gmax: float = DEFAULT_GMAX,
-    smax: float = DEFAULT_SMAX,
     pregrad: str | None = None,
     postgrad: str | None = None,
     version: float = 5,
+    acq: Acquisition | None = None,
     **kwargs,
 ):
     """Calculate gradients from k-space points and write to file.
@@ -229,28 +220,14 @@ def write_trajectory(
     trajectory : np.ndarray
         Trajectory in k-space points.
         Shape (num_shots, num_samples_per_shot, dimension).
-    FOV : tuple
-        Field of view
-    img_size : tuple
-        Image size
     grad_filename : str
         Gradient filename
-    norm_factor : float, optional
-        Trajectory normalization factor, by default 0.5
-    gamma : float, optional
-        Gyromagnetic ratio in kHz/T, by default 42.576e3
-    raster_time : float, optional
-        Gradient raster time in ms, by default 0.01
     check_constraints : bool, optional
         Check scanner constraints, by default True
     TE_pos : float, optional
         The ratio of trajectory when TE occurs, with 0 as start of
         trajectory and 1 as end. By default 0.5, which is the
         center of the trajectory (in-out trajectory).
-    gmax : float, optional
-        Maximum gradient magnitude in T/m, by default 0.04
-    smax : float, optional
-        Maximum slew rate in T/m/ms, by default 0.1
     pregrad : str, optional
         Pregrad method, by default `prephase`
         `prephase` will add a prephasing gradient to the start of the trajectory.
@@ -272,20 +249,11 @@ def write_trajectory(
     # Convert normalized trajectory to gradients
     gradients, initial_positions, final_positions = convert_trajectory_to_gradients(
         trajectory,
-        norm_factor=norm_factor,
-        resolution=np.asarray(FOV) / np.asarray(img_size),
-        raster_time=raster_time,
-        gamma=gamma,
+        acq,
         get_final_positions=True,
     )
     Ns_to_skip_at_start = 0
     Ns_to_skip_at_end = 0
-    scan_consts = {
-        "gamma": gamma,
-        "gmax": gmax,
-        "smax": smax,
-        "raster_time": raster_time,
-    }
     if pregrad == "prephase":
         if version < 5.1:
             raise ValueError(
@@ -295,7 +263,7 @@ def write_trajectory(
         start_gradients = get_gradient_amplitudes_to_travel_for_set_time(
             kspace_end_loc=initial_positions,
             end_gradients=gradients[:, 0],
-            **scan_consts,
+            acq=acq,
         )
         initial_positions = np.zeros_like(initial_positions)
         gradients = np.hstack([start_gradients, gradients])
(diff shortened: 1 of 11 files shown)


src/mrinufft/io/nsp.py
```diff
@@ -207,19 +204,13 @@ def _pop_elements(array, num_elements=1, type=np.float32):
...
-    img_size: tuple[int, ...],
     grad_filename: str,
-    norm_factor: float = KMAX,
-    gamma: float = Gammas.HYDROGEN,
-    raster_time: float = DEFAULT_RASTER_TIME,
     check_constraints: bool = True,
     TE_pos: float = 0.5,
-    gmax: float = DEFAULT_GMAX,
-    smax: float = DEFAULT_SMAX,
     pregrad: str | None = None,
```
chaithyagr, line 210: These codes are kinda used in legacy codebases (even SPARKLING). I feel its good practice to have both options available for now and add a warning. WDYT?

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,220 @@ class VDSpdf(StrEnum):
...
+        Image size in pixels (x, y, z).
+    hardware : Hardware
+        Hardware configuration for the acquisition.
+    gamma : Gammas, optional
+        Gyromagnetic ratio in Hz/T for the nucleus being imaged.
+        Defaults to `Gammas.HYDROGEN`.
+    oversampling : int, optional
+        Oversampling factor for the ADC. Defaults to 1.
+    norm_factor : float, optional
+        Normalization factor for the trajectory. Defaults to 0.5.
```
chaithyagr, line 289: I believe this isnt strictly a part of "acquisition" config. But I am okay let this pass as it is quite close

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,220 @@ class VDSpdf(StrEnum):
...
     """
-    return trajectory * norm_factor * (2 * resolution)
+    acq = acq or Acquisition.default
+    return trajectory * acq.norm_factor * (2 * acq.res)
 
 
 def unnormalize_trajectory(
     trajectory: NDArray,
-    norm_factor: float = KMAX,
-    resolution: float | NDArray = DEFAULT_RESOLUTION,
```
chaithyagr, line 210: Im not a fan of needing to provide all the acquisition config here. We should also be able to just specify the resolution rather. Right?

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,220 @@ class VDSpdf(StrEnum):
...
+
+    def __getattr__(self, name):
+        """Allow access to hardware attributes directly from Acquisition."""
+        return getattr(self.hardware, name)
+
+    @property
+    def res(self) -> tuple[float, ...]:
+        """Resolution in meters."""
+        return tuple(fov / size for fov, size in zip(self.fov, self.img_size))
+
```
chaithyagr, line 333: Another property:

def kmax():


This returns the kmax for this trajectory

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,220 @@ class VDSpdf(StrEnum):
...
+    acquisition configuration without needing to instantiate a new `Acquisition` object
+    each time.
+
+    """
+
+    default: ClassVar[Acquisition]
+
+    fov: tuple[float, float, float]  # Field of View in m
+    img_size: tuple[int, int, int]  # Image size in pixels
+    hardware: Hardware = SIEMENS.TERRA  # Hardware configuration
```
chaithyagr, line 315: I think having hardware inside Acquisition could over-complicate things. We dont always need a hardware config really, only when we specify or check for hardware constraints

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,220 @@ class VDSpdf(StrEnum):
...
+        Field of View in meters (x, y, z).
+    img_size : tuple[int, int, int]
+        Image size in pixels (x, y, z).
+    hardware : Hardware
+        Hardware configuration for the acquisition.
+    gamma : Gammas, optional
+        Gyromagnetic ratio in Hz/T for the nucleus being imaged.
+        Defaults to `Gammas.HYDROGEN`.
+    oversampling : int, optional
+        Oversampling factor for the ADC. Defaults to 1.
```
chaithyagr, line 287: Not really needed if you have dwell time in hardware config. Also, I feel dwell time should be here rather than in "hardware config", as it is not specific to a scanner type but rather a scan itself.

src/mrinufft/trajectories/utils.py
chaithyagr: To me we can keep hardware and acquisition as 2 different dataclasses and not nest them.

chaithyagr (changes requested):
Thank you for this amazing work, First round of reviews with some broader questiosn. Will do a review again soon

src/mrinufft/trajectories/utils.py
paquiteau: good idea

src/mrinufft/trajectories/utils.py
paquiteau: I disagree. An acquisition is always performed on some specific (or a generic preset) hardware. Having a single structure to pass by simplifies everything.

src/mrinufft/trajectories/utils.py
paquiteau: Well the constructor could either take oversampling or dwell_time at init, we have a similar issue with resolution as well. I think some getter/setter wizardry could do the trick.

src/mrinufft/trajectories/utils.py
chaithyagr: I agree about simplification. But i still feel acquisition and hardware are fundamentally different things. Within this PR itself i see the same dataclass being sent even if we use only one of them. I'm ok with it for simplification but it's better to be mindful that we potentially make it harder for users to just give hardware config to check constraints etc...

src/mrinufft/trajectories/utils.py
chaithyagr: The thing is oversampling was originally defined from sparkling, but i want to get away from its use, we just sample at higher rate, it's not "over" sampled , which generally leads people to believe that you are above nyquist rate by that factor

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,268 @@ class VDSpdf(StrEnum):
...
```
chaithyagr, line 287: Docs need to be changed still right?

src/mrinufft/trajectories/utils.py
```diff
@@ -173,69 +177,268 @@ class VDSpdf(StrEnum):
...
+    acquisition configuration without needing to instantiate a new `Acquisition` object
+    each time.
+
+    """
+
+    fov: tuple[float, float, float]  # Field of View in m
+    img_size: tuple[int, int, int]  # Image size in pixels
+    hardware: Hardware = SIEMENS.TERRA  # Hardware configuration
+    gamma: Gammas = Gammas.HYDROGEN  # Hz/T
+    adc_dwell_time: float = 1 * SI.micro  # us
```
chaithyagr, line 316: 1 us is too low, lets start with 5us as default.

chaithyagr:
I would prefer to also see docs, can you make you merge with commit message [docs] to trigger a docs build?

Issue #289, opened 2025-09-08
Add gpu_device_id argument to cufinufft and gpunufft operators


AsmaTANABEN:
## Description & use cases

To support multi-GPU setups and distributed training, cufinufft and gpunufft operators should accept a `gpu_device_id` argument. This enables to set the cuda device before any gpu resources are initialized.

Currently, these operators use by default gpu device 0, limiting parallelism.

## Suggested implementation

* Add `gpu_device_id` as an optional argument in kwargs.
* In cufinufft, set the cuda device in CuPy early in the constructor using `cp.cuda.Device(gpu_device_id).use()`.
* Pass `gpu_device_id` to the gpunufft operator’s NUFFTOp arguments, using the modified gpunufft code (PR opened).
* Handle invalid device id with warnings.

PR #290, opened 2025-09-08
Added support for multi-GPU runs with extra optional  in kwargs for both cufinufft and gpunufft operators

AsmaTANABEN:
- Solves #289

This PR adds support for selecting the CUDA device using `gpu_device_id` in both `cufinufft` and `gpunufft` operators:
-  Adds optional `gpu_device_id` to kwargs
- Sets cupy cuda device in cufinufft constructor.
- Passes `gpu_device_id` to gpunufft's NUFFTOp (compatible with recent changes in gpuNUFFT https://github.com/chaithyagr/gpuNUFFT/pull/26)

Usage example: 
<pre>``` import mrinufft 
import numpy as np
nufft = mrinufft.get_operator(
    backend_name="cufinufft", wrt_data=True, wrt_traj=False
)(
    samples=np.random.rand(100,3),
    shape=(20,20,20),
    density="pipe",
    squeeze_dims=False,
    gpu_device_id=1
)``` </pre>
If the specified gpu_device_id is invalid, a warning will be issued like:
`Failed to set CUDA device 1: cudaErrorInvalidDevice: invalid device ordinal`
and the code will fall back to CUDA device 0. Otherwise, device 1 will be used.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 
-gpunufft = ["gpuNUFFT>=0.9.0", "cupy-cuda12x"]
+gpunufft = ["gpuNUFFT>=0.10.1", "cupy-cuda12x"]
 
 torchkbnufft = ["torchkbnufft", "cupy-cuda12x"]
 torchkbnufft-cpu = ["torchkbnufft", "cupy-cuda12x"]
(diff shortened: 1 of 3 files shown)


chaithyagr:
Ideally, you should be able to "fake it till u make it" :P 

Basically, set `export CUDA_VISIBLE_DEVICES=0,0`
And inside python:

`torch.cuda.device_count = lambda : 2`


This sorta fakes pytorch to think there are 2 GPUs and both are mapped to GPU 0.

PR #295, opened 2025-09-19
Upgrade to python 3.10 and minimum dependencies version

paquiteau:
- Closes None

- Set minimum version for our dependencies, to ensure proper resolving 
- The required python version is now 3.10 (3.9 EOL was/is october 2025)

feat: set minimum versions for dependencies.

### DIFF
--- a/src/mrinufft/trajectories/tools.py
+++ b/src/mrinufft/trajectories/tools.py
@@ -1,6 +1,7 @@
 """Functions to manipulate/modify trajectories."""
 
-from typing import Any, Callable, Literal, Optional
+from typing import Any, Literal, Optional
+from collections.abc import Callable
 
 import numpy as np
 from numpy.typing import NDArray
@@ -417,10 +418,10 @@ def _plateau_value(gs, ge, n_down, n_up, n_pl, area_needed):
 
 
 def get_gradient_times_to_travel(
-    kspace_end_loc: Optional[NDArray] = None,
-    kspace_start_loc: Optional[NDArray] = None,
-    end_gradients: Optional[NDArray] = None,
-    start_gradients: Optional[NDArray] = None,
+    kspace_end_loc: NDArray | None = None,
+    kspace_start_loc: NDArray | None = None,
+    end_gradients: NDArray | None = None,
+    start_gradients: NDArray | None = None,
     gamma: float = Gammas.Hydrogen,
     raster_time: float = DEFAULT_RASTER_TIME,
     gmax: float = DEFAULT_GMAX,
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -18,7 +18,7 @@ torchkbnufft-cpu = ["torchkbnufft", "cupy-cuda12x"]
 torchkbnufft-gpu = ["torchkbnufft", "cupy-cuda12x"]
 
 cufinufft = ["cufinufft>=2.4.0", "cupy-cuda12x"]
-tensorflow = ["tensorflow-mri==0.21.0", "tensorflow-probability==0.17.0", "tensorflow-io==0.27.0", "matplotlib==3.7"]
+tensorflow = ["tensorflow-mri>=0.22.0"]
```
chaithyagr, line 21: Did you test this? I remember this versioning was very specific to avoid issues

src/mrinufft/trajectories/tools.py
```diff
@@ -417,10 +418,10 @@ def _plateau_value(gs, ge, n_down, n_up, n_pl, area_needed):
...
 
 def get_gradient_times_to_travel(
-    kspace_end_loc: Optional[NDArray] = None,
-    kspace_start_loc: Optional[NDArray] = None,
-    end_gradients: Optional[NDArray] = None,
-    start_gradients: Optional[NDArray] = None,
+    kspace_end_loc: NDArray | None = None,
+    kspace_start_loc: NDArray | None = None,
+    end_gradients: NDArray | None = None,
+    start_gradients: NDArray | None = None,
```
chaithyagr, line 424: Why though? Is this a new way to type?

src/mrinufft/trajectories/tools.py
```diff
@@ -524,10 +525,10 @@ def _residual(gi):
 
 def get_gradient_amplitudes_to_travel_for_set_time(
     kspace_end_loc: NDArray,
-    kspace_start_loc: Optional[NDArray] = None,
-    end_gradients: Optional[NDArray] = None,
-    start_gradients: Optional[NDArray] = None,
-    nb_raster_points: Optional[int] = None,
```
chaithyagr, line 530: Cause my linter fails, it underlines possible type miss match errors

chaithyagr (commented):
Thank you! Most look good to me. I just have some questions, we don't need a second review, just answer them and once tests pass we can merge

src/mrinufft/trajectories/tools.py
paquiteau: Its totally equivalent, PEP 604 ;)

chaithyagr:
I guess i need a better linter

pyproject.toml
paquiteau: Not really, but tensorflow-mri manages its own dependency internally. Its also removes the pinning on matplotlib that was causing issue with numpy 2.0

PR #299, opened 2025-09-24
Lint and house keeping

### LINKED ISSUE #235: Comment the API docs link from artifacts using a bot in case we build the docs
This helps in allowing reviewers to also review the docs.

paquiteau:
- Closes #235 

- Add more type annotation and cleanup 
- some magic is starting to happens in the docstring 6695b48a6134b1ae5d3dd9b55ab9186918809bc3 c63fa9589c2d6f96af91e400479cc61f02eba52d 

I tried to keep a nice commit history, so that we can rebase directly


paquiteau:
@chaithyagr check this out

/build-docs

pyproject.toml
```diff
@@ -28,7 +28,6 @@ pynufft-cpu = ["pynufft"]
 pynufft-gpu = ["pynufft"]
 
 io = ["pymapvbvd"]
-smaps = ["scikit-image"]
```
chaithyagr, line 31: Why though?

src/mrinufft/_array_compat.py
chaithyagr, line 1: I dont see this in docs? Where is it?

pyproject.toml
paquiteau: smaps estimation is part of the extra module, which already has scikit-image as dependency, so let's just make things simple and reduce the number of possibility to install.

src/mrinufft/_array_compat.py
paquiteau: It is not exposed (afterall it was a private module), however I have recents commit that exposes it. I will push them soon.

chaithyagr: Are they in this PR? Lemme know if you want me to review again

PR #300, opened 2025-09-24
feat: rework ORC wrapper and estimations.

### LINKED ISSUE #191: SVD-based off-resonance correction coefficients
As discussed in PR #180 we should provide at some point the SVD-based coefficients for off-resonance correction [1]. As a reminder, the SVD-based solution is the optimal solution to the matrix factorization problem considered for off-resonance correction.

[1] Fessler, Jeffrey A., Sangwoo Lee, Valur T. Olafsson, Hugo R. Shi, and Douglas C. Noll. "Toeplitz-based iterative image reconstruction for MRI with correction for magnetic field inhomogeneity." IEEE Transactions on Signal Processing 53, no. 9 (2005): 3393-3402.

### LINKED ISSUE #192: Automatic number of bins & interpolators for off-resonance correction
As discussed in PR #180, we should provide at some point an option in the off-resonance correction operator to automatically suggest a number of bins and interpolators for a given B0+R2* field map. The default options should also be based on their error in matrix factorization in the case of a typical 3T or 7T B0 field map.

### LINKED ISSUE #254: Off-resonance correction works in opposite convention

paquiteau:
- Closes #254: (Just negate the field map on your side if needed)
- Closes #191
- Closes #192 

- Built on top of #299 

- :flight_arrival: incorporate code from pysap-mri (SVD-based interpolation)
- :broom:  Cleanup the instantiation of the ORC wrapper 
- :sparkles: Add a Lazy Array to avoid too much memory for the space interpolator when possible 
- :magic_wand: Showcase the docstring template update proposed in #299

### DIFF
--- a/tests/operators/test_orc.py
+++ b/tests/operators/test_orc.py
@@ -84,7 +84,7 @@ def test_orc_forward(
     ksp = orc_nufft.op(image_data)
     ksp_ideal = ext_mat @ image_data.ravel()
 
-    assert_correlate(ksp.squeeze(), ksp_ideal)
+    assert_correlate(ksp.squeeze(), ksp_ideal, slope_err=0.005)
 
 
 def test_orc_adjoint(orc_info, kspace_data):
@@ -93,4 +93,4 @@ def test_orc_adjoint(orc_info, kspace_data):
     img = orc_nufft.adj_op(kspace_data)
     img_ideal = ext_mat.conj().T @ kspace_data.ravel()
     img_ideal = img_ideal.reshape(orc_nufft.shape)
-    assert_correlate(img.squeeze(), img_ideal)
+    assert_correlate(img.squeeze(), img_ideal, slope_err=0.005)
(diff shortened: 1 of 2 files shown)


Daval-G (commented):
Quite elegant, I don't have much to say about this MR for now (I assume you will add more things, notably to address #192). Thanks for adding interpolators from pysap-mri, it's a great addition to have multiple methods from the literature (and also SVD is optimal for our usage so...)

paquiteau:
Just a heads up on the status:
- [x] Added MFI, MTI and SVD methods for computing the B,C interpolators (with both B0-only or B0+R2* mapping)
- [X] Refactor and cleanup more stuff
- [ ] Test rewrites
- [ ] Example rewrites

PR #316, opened 2025-10-23
Add LSQR and LSMR for solving the pseudo inverse

### LINKED ISSUE #216: Conversion as LinearOperator from scipy
The `LinearOperator` interface [^1] is proposed by Scipy for describing linearOperator which are too big to be represented as matrices. MRI-NUFFT Operators are linear operators, so we could support quite easily. 
`LinearOperator` are also used in PyLops [^2] (and PyProximal [^3]). 

In term of implementation I think a wrapper class, and a custom method (`FourierBaseOperator.make_linear_operator`) should do the trick.  

[^1]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html
[^2]: https://pylops.readthedocs.io/en/latest/#
[^3]: https://pyproximal.readthedocs.io/en/stable/index.html

### LINKED ISSUE #296: Add Thikhonov Regularization for CG
## Description & use cases

> Describe the feature you would like to see in MRI-nufft. Be as specific as possible about what you want to achieve.

Conjugate gradient descent minimize the least square problem $\|Ax-y\|_2^2$ It can also solve the generalization \|Ax-y\|_2^2 + \lambda \|x\|_2^2$. 
## Suggested implementation

> Describe how you envision the feature being implemented. This could include specific functions, classes, or changes to existing code.

Modify the cg function to enable this. Also check for tweaks in the various operators.

### LINKED ISSUE #307: Homogeneize number of iterations parameters
## Description & use cases

> Describe the feature you would like to see in MRI-nufft. Be as specific as possible about what you want to achieve.

- All over mrinufft we have iterative methods that relies on a given number of iterations, but we have no constistent naming (`max_iter`, `num_iterations, `n_iter`, etc...) 

## Suggested implementation

> Describe how you envision the feature being implemented. This could include specific functions, classes, or changes to existing code.

- just use `n_iter` everywhere :)

paquiteau:
- Closes #296
- Closes #216
- Closes #307 
- 

- Add the LSQR solver, which is 20% faster per iteration, and stop earlier than the current CG optimizer
- change the `.cg` to `.pinv_solver` for the NUFFT operator.
- Add Tikhonov regularization / ridge regression. 

NB: since we solve the general least-square problem $\|Ax-b\|_2^2_ + \gamma \|x-x_0\|_2^2$ we also get the proximal operator implementation for free (Useful for HQS schemes)

### DIFF
--- a/src/mrinufft/operators/interfaces/cufinufft.py
+++ b/src/mrinufft/operators/interfaces/cufinufft.py
@@ -860,7 +860,7 @@ class MRICufiNUFFT(FourierOperatorBase):
         cls,
         kspace_loc,
         volume_shape,
-        num_iterations=10,
+        max_iter=10,
         osf=2,
         normalize=True,
         **kwargs,
@@ -873,7 +873,7 @@ class MRICufiNUFFT(FourierOperatorBase):
             the kspace locations
         volume_shape: np.ndarray
             the volume shape
-        num_iterations: int default 10
+        max_iter: int default 10
             the number of iterations for density estimation
         osf: float or int
             The oversampling factor the volume shape
@@ -894,7 +894,7 @@ class MRICufiNUFFT(FourierOperatorBase):
             **kwargs,
         )
         density_comp = cp.ones(kspace_loc.shape[0], dtype=grid_op.cpx_dtype)
-        for _ in range(num_iterations):
+        for _ in range(max_iter):
             density_comp /= cp.abs(
                 grid_op.op(
                     grid_op.adj_op(density_comp.astype(grid_op.cpx_dtype))
--- a/src/mrinufft/operators/interfaces/finufft.py
+++ b/src/mrinufft/operators/interfaces/finufft.py
@@ -174,7 +174,7 @@ class MRIfinufft(FourierOperatorCPU):
         cls,
         kspace_loc,
         volume_shape,
-        num_iterations=10,
+        max_iter=10,
         osf=2,
         normalize=True,
         **kwargs,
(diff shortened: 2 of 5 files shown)


src/mrinufft/extras/optim.py
```diff
@@ -0,0 +1,763 @@
...
+    base_params=r"""
+nufft: FourierOperatorBase
+    The NUFFT operator representing the forward model.
+kspace_data: NDArray
+    The right-hand side vector (`kspace` data). Shape typically (n_batchs,
+    n_coils, n_samples).
+damp: float, optional
+    Damping (regularization) parameter. Default is 0.0 (no regularization).
+x0: NDArray or None, optional
+    Damping vector. If None, uses zero. Shape must be
```
chaithyagr: Shape must be ? I think something is missing here

src/mrinufft/extras/optim.py
```diff
@@ -0,0 +1,763 @@
...
+    operator: FourierOperatorBase,
+    kspace_data: NDArray,
+    damp: float = 0.0,
+    x0: NDArray | None = None,
+):
+    """
+    Compute the regularized least squares loss for MRI reconstruction.
+
+    Computes the loss:
+        ||A x - y||_2^2 + damp^2 ||x - x0||_2^2
```
chaithyagr: Use :math ?

src/mrinufft/extras/optim.py
chaithyagr: On a general note, I am curious what is the advantage of having our own implementation rather than just getting LinearOperator and expecting cupy / numpy to do the iterations for us?

chaithyagr (changes requested):
Almost there, great work! I don't think there's another review needed, please merge when green and after addressing the issues

PR #319, opened 2025-11-04
Add support for ESPIRIT based SMaps (verrry efficient :P )

### LINKED ISSUE #305: Fix Smaps estimation and add support for ESPiRIT
While sigpy has its implementations, its usually very memory hungry.
It can help to have memory efficient implementations for our specific use cases. Also, it can help to couple it with non-Cartesian MRI, by leveraging the extract_center and doing a DC adjoint to get per channel low res image.

Also, it might help fixing the `low-frequency` method to remove ringing issues which is commonly seen due to bad sinc interpolation with NUFFT

chaithyagr:
- Closes #305 

I have added support for ESPIRIT algorithm inspired by sigpy version of the codes. However, I added some really helpful codes to allow for memory efficiency and made sure API is similar to ours. 

- A new example which showcases different Smap estimation methods, that too in 3D

Here are current results from example (although i think we need to look into docs anyway):

I simulated the following Smaps:

image

On following brainweb-dl data:

image

I estimate smaps Using ESPIRIT:

image

and using our earlier `low_frequency`:

image

TODO:
- Add support for pykeops based implementation of AHA for more memory efficiency, but to be discussed based on user feedback.

### DIFF
--- a/examples/extras/example_smaps.py
+++ b/examples/extras/example_smaps.py
@@ -51,7 +51,7 @@ def show_maps(imgs):
     return fig
 
 
-BACKEND = os.environ.get("MRINUFFT_BACKEND", "gpunufft")
+BACKEND = os.environ.get("MRINUFFT_BACKEND", "cufinufft")
 
 # %%
 # Get MRI data, 3D FLORET trajectory, and simulate k-space data
@@ -79,7 +79,7 @@ forward_op = get_operator(BACKEND)(
 )
 kspace_data = forward_op.op(per_ch_mri)  # Simulate k-space data
 
-if BACKEND == "gpunufft":
+if str.lower(BACKEND) in ["gpunufft", "cufinufft"]:
     # GPU exists, run on GPU
     import cupy as cp
 
--- a/src/mrinufft/extras/smaps.py
+++ b/src/mrinufft/extras/smaps.py
@@ -75,7 +75,7 @@ def _extract_kspace_center(
             )
         )
         center_locations = kspace_loc[condition, :]
-        data_thresholded = kspace_data[:, condition]
+        data_thresholded = kspace_data[..., condition]
         if density is not None:
             dc = density[condition]
         else:
@@ -259,11 +259,11 @@ def low_frequency(
     return Smaps
 
 
-@with_torch
 def _unfold_blocks(calib, calib_width):
-    for i, width in enumerate(calib_width):
-        calib = calib.unfold(dimension=i + 1, size=width, step=1)
-    return calib
+    xp = get_array_module(calib)
+    return xp.lib.stride_tricks.sliding_window_view(
+        calib, calib_width, axis=tuple(range(1, calib.ndim))
+    )
 
 
 @register_smaps
--- a/src/mrinufft/extras/optim.py
+++ b/src/mrinufft/extras/optim.py
@@ -777,7 +777,7 @@ def cg(
     """
     lipschitz_cst = operator.get_lipschitz_cst()
     if operator.backend == "cufinufft":
-        lipschitz_cst = lipschitz_cst.get()
+        lipschitz_cst = float(lipschitz_cst.get())
     xp = get_array_module(kspace_data)
     image = (
         xp.zeros(operator.img_full_shape, dtype=kspace_data.dtype)
--- a/src/mrinufft/operators/base.py
+++ b/src/mrinufft/operators/base.py
@@ -304,7 +304,7 @@ class FourierOperatorBase(ABC):
             method = get_smaps(method)
         if not isinstance(method, Callable):
             raise ValueError(f"Unknown smaps method: {method}")
-        smaps, SOS = method(
+        smaps = method(
             self.samples,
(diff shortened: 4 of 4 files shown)


examples/extras/example_smaps.py
```diff
@@ -0,0 +1,93 @@
...
+        extra_kwargs["decim"] = 4
+    Smaps = get_smaps(method)(
+        samples_loc,
+        mri.shape,
+        kspace_data=kspace_data,
+        density=forward_op.density,
+        backend=BACKEND,
+        **extra_kwargs,
+    )
+    show_maps(Smaps.get())
```
paquiteau: I would simply unroll the loop here. Make it easier to read/understand, can also be the opportunity to details a bit more about the arguments.

src/mrinufft/extras/cartesian.py
```diff
@@ -0,0 +1,33 @@
+"""Cartesian FFT and iFFT utilities."""
+
+from mrinufft._array_compat import get_array_module
+
+
+def fft(image, dim=3, shape=None):
+    """n-dimensional FFT along the last dim axes."""
+    axes = range(-dim, 0)
```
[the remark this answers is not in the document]
paquiteau: A see also to numpy fft could be nice as well, for further explanation on the arguments.

src/mrinufft/extras/cartesian.py
```diff
@@ -0,0 +1,33 @@
...
+            norm="ortho",
+            axes=axes,
+            s=shape,
+        ),
+        axes=axes,
+    )
+
+
+def ifft(kspace, dim=3, shape=None):
+    """n-dimensional inverse FFT along the last dim axes."""
```
paquiteau:
```suggestion
def ifft(kspace, dims=3, shape=None):
    """n-dimensional inverse FFT along the last ``dims`` axes."""
```
+ the same change as in FFT.

src/mrinufft/extras/smaps.py
```diff
@@ -194,18 +238,181 @@ def low_frequency(
...
+    calib = _crop_or_pad(central_kspace, calib_shape)
+    calib = _unfold_blocks(calib, kernel_width)
+    calib = calib.reshape(
+        calib.shape[0],
+        -1,
+        np.prod(kernel_width),
+    )
+    calib = calib.transpose(1, 0, 2).reshape(-1, calib.shape[0] * np.prod(kernel_width))
+    _, S, VH = xp.linalg.svd(calib, full_matrices=False)
+    VH = VH[S > thresh * S.max(), :]
```
paquiteau: If this is a bottleneck, something along these lines could help: 
```python
    max_rank = np.min(calib.shape) -1
    k = np.min(max_rank, 10)
    svds = sp.linalg.sparse.svds if xp == np  else cupyx.scipy.sparse.linalg.svds
    while k < max_rank:
       _, S, VH = svds(calib,k=k full_matrices=False)
       if min(S) < max(S)*thresh: 
           break
       else: 
           k = np.min(max_rank, k+5)
    VH = VH[S > thresh * S.max(), :]
```

src/mrinufft/_utils.py
paquiteau: so now the trajectory is always on cpu ? this is not so good for cufinufft (which want to have it on gpu)

src/mrinufft/extras/smaps.py
```diff
@@ -194,18 +238,181 @@ def low_frequency(
...
+        SOS = np.linalg.norm(Smaps, axis=0) + 1e-10
+    Smaps = Smaps / SOS
+    return Smaps
+
+
+@with_torch
+def _unfold_blocks(calib, calib_width):
+    for i, width in enumerate(calib_width):
+        calib = calib.unfold(dimension=i + 1, size=width, step=1)
+    return calib
```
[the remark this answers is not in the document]
chaithyagr: Also need it with GPU (on cupy :P )
Also, we really need to create a copy anyway to do SVD

src/mrinufft/extras/smaps.py
chaithyagr: Is this better? Especially as we need all the eigen vectors in the end anyway

paquiteau: it might if you have a very low rank matrix

src/mrinufft/extras/smaps.py
```diff
@@ -194,18 +246,245 @@ def low_frequency(
...
+    Smaps : NDArray
+        The sensitivity maps
+
+    References
+    ----------
+    Uecker M, Lai P, Murphy MJ, Virtue P, Elad M, Pauly JM, Vasanawala SS,
+    Lustig M. ESPIRiT--an eigenvalue approach to autocalibrating parallel
+    MRI: where SENSE meets GRAPPA. Magn Reson Med.
+    2014 Mar;71(3):990-1001. doi: 10.1002/mrm.24751.
+    PMID: 23649942; PMCID: PMC4142121.
```
paquiteau: I think the doc formatting is wrong here

PR #325, opened 2025-11-12
More docs improvements

paquiteau:
- Closes None

- [x] Provide a documentations for the `get_xyz` functions that was lacking (using a custom sphinx directive to collect all registered functions in it)
- [x] link to code on github instead of local rst file
- [x] Deploy to github actions rather than gh-pages, for lightweight repo clone (but still with support for 'run on colab' examples)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,7 +163,7 @@ Don't hesitate to look for unsolved `issues <https://github.com/mind-inria/mri-n
 .. _finufft: https://github.com/flatironinstitute/finufft
 
 .. |gpunufft| replace:: ``gpunufft``
-.. _gpunufft: https://github.com/chaihtyagr/gpunufft
+.. _gpunufft: https://github.com/chaithyagr/gpunufft
 
 .. |torchkbnufft| replace:: ``torchkbnufft``
 .. _torchkbnufft: https://github.com/mmuckley/torchkbnufft


.github/workflows/docs.yml
```diff
@@ -44,6 +44,15 @@ jobs:
             });
             core.setOutput('head_ref', pr.head.ref);
 
+      - name: Comment That docs building is starting
```
chaithyagr: Can we add back the check for [docs] in commit string please? It helps for folks making PR from forks.

.github/workflows/docs.yml
```diff
@@ -74,6 +83,14 @@ jobs:
           export MRINUFFT_BACKEND=cufinufft
           python -m sphinx docs docs_build ${{ contains(github.event.comment.body, 'fast') && '-D plot_gallery=0' || '' }} -j 5
 
+
+      - name: Upload artifact for pages
+        if: ${{ github.event_name == 'workflow_run' }}
+        uses: actions/upload-pages-artifact@v4
+        with:
+          path: docs_build
```
chaithyagr: Ill take your word that it works fine :)

docs/getting_started.rst
```diff
@@ -4,27 +4,64 @@ Getting Started
...
 Development Version
 ~~~~~~~~~~~~~~~~~~~
 
-If you want to modifiy the mri-nufft code base
+If you want to modify the mri-nufft code base
+
+.. code-block:: sh
+
+    git clone https://github.com:mind-inria/mri-nufft
+    pip install -e ./mri-nufft[dev,doc,extra,io,autodiff,tests,cufinufft,gpunufft,finufft]
```
chaithyagr: Cant we make one for having it all? Also for dev do we need all?

docs/link_info.py
```diff
@@ -0,0 +1,97 @@
+"""Sphinx linkcode_resolve function to link to GitHub source code."""
```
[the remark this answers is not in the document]
paquiteau: Sphinx linkcode is an extension that get a correct link to the function definition on github. It's mostly cargo culting from stackoverflow.

PR #329, opened 2025-11-18
allow external progress bar to be used

paquiteau:
Externally defined progress bar can be passed to methods that requires them (typically pinv_solver), by reusing them over call, it reduces the std output, and makes it more manageable. 

- Closes None

- makes possible to use externally defined progress bar from tqdm
-

feat: pass existing progressbars

### DIFF
--- a/src/mrinufft/extras/optim.py
+++ b/src/mrinufft/extras/optim.py
@@ -4,13 +4,14 @@
 """
 
 from collections.abc import Callable
-import numpy as np
-from tqdm.auto import tqdm
-from numpy.typing import NDArray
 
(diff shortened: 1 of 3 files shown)


src/mrinufft/extras/optim.py
```diff
@@ -5,12 +5,12 @@
 
 from collections.abc import Callable
 import numpy as np
-from tqdm.auto import tqdm
 from numpy.typing import NDArray
+from tqdm.auto import tqdm
```
chaithyagr: Thats unnecessary change?

chaithyagr (changes requested):
Minor change, otherwise LGTM

PR #338, opened 2025-12-14
Add: enable tensor trajectory initialization and updates

### LINKED ISSUE #147: Prevent duplications of `samples`, esp when using with autodiff
Currently we seem to have 2 different memory elements for `samples` (perhaps 3 assuming interfaces has a copy of its own internally, but that cant be fixed).

What we need is to ensure that we support samples to be on CPU or GPU and handle accordingly.

JoosenLi:
- Closes #

-Enable tensor-based (GPU/CPU) initialization and updates of trajectories.
(You can still use NumPy if preferred, but this will introduce additional CPU–GPU data transfers during neural network training, increasing runtime.)

### DIFF
--- a/src/mrinufft/operators/base.py
+++ b/src/mrinufft/operators/base.py
@@ -697,6 +697,29 @@ class FourierOperatorBase(ABC):
 
     @samples.setter
     def samples(self, new_samples: NDArray[np.floating]):
+        self.update_samples(new_samples)
+
+    def update_samples(
+        self, new_samples: NDArray[np.floating], *, unsafe: bool = False
+    ):
+        """Update the samples of the NUFFT operator.
+
+        Parameters
+        ----------
+        new_samples: np.ndarray or GPUArray
+            The new samples location of shape ``Nsamples x N_dimensions``.
+        unsafe: bool, default False
+            If True, the original array is used directly without any checks.
+            This should be used with caution as it might lead to unexpected behavior.
+
+        Notes
+        -----
+        If unsafe is True, the new_samples should be of shape (Nsamples, N_dimensions),
+        F-ordered (column-major) and in the range [-pi, pi]. If not, this will lead to
+        unexpected behavior. You have been warned.
+
+        If unsafe is False, this is automatically handled.
+        """
         self._samples = new_samples
 
     @property
--- a/src/mrinufft/operators/interfaces/cufinufft.py
+++ b/src/mrinufft/operators/interfaces/cufinufft.py
@@ -23,6 +23,7 @@ CUFINUFFT_AVAILABLE = CUPY_AVAILABLE
 try:
     import cupy as cp
     from cufinufft import Plan
+    from cufinufft._compat import get_array_ptr
 except ImportError:
     CUFINUFFT_AVAILABLE = False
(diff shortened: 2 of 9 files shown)


paquiteau:
Thanks for the PR! A few remarks:
- You can make this much simpler by relying on the with_numpy_cupy decorator, that would handles the conversion to numpy or cupy (from torch or any other library) before calling the function. I think it is needed for proper_trajectory and the sample setter as well. (this removes the need for your `_process_samples` method)
- Overall, there should not be any call to torch in `cufinufft.py`, as it is an optional dependency, we should be able to do with cupy with only a limited amount of copies (be sure that the trajectories are normalized to -pi/pi and in column major/fortran order)
- Could you add a test case for this ? ttypically,by augmenting the `test_update.py::test_op/adj_op_samples` test ? by also testing support for on-device gpu for cufinufft operators only.

JoosenLi:
Hi @paquiteau,

I was going through the profiles in "samples_update.py" you provided and noticed a small discrepancy in line 114. 

The code calls `proper_trajectory(traj_cp, order="F", ...)` but it seems the `proper_trajectory()` function definition doesn't actually accept an 'order' keyword argument. This would trigger a TypeError during execution.

paquiteau:
ah yes, that was from earlier testing, I ended up removing this argument, so you can discard it as well :D

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -212,9 +222,9 @@ def __init__(
         self.squeeze_dims = squeeze_dims
         self.n_coils = n_coils
         self.autograd_available = True
-        # For now only single precision is supported
-        self._samples = np.asfortranarray(
-            proper_trajectory(samples, normalize="pi").astype(np.float32, copy=False)
+        self._samples = proper_trajectory(samples, normalize="pi")
+        self._samples = cp.array(
+            proper_trajectory(samples, normalize="pi"), order="F", copy=False
```
chaithyagr: Why call proper trajectory twice?

src/mrinufft/operators/base.py
```diff
@@ -691,6 +691,29 @@ def samples(self) -> NDArray:
...
+            If True, the original array is used directly without any checks.
+            This should be used with caution as it might lead to unexpected behavior.
+
+        Notes
+        -----
+        If unsafe is True, the new_samples should be of shape (Nsamples, N_dimensions),
+        F-ordered (column-major) and in the range [-pi, pi]. If not, this will lead to
+        unexpected behavior. You have been warned.
+
+        If unsafe is False, this is automatically handled.
```
chaithyagr: Isnt this doc like very much repeated? Can we have a single string and add it everywhere?

chaithyagr (changes requested):
Thank you for these wonderful changes and perf fixes. Some minor comments

src/mrinufft/operators/interfaces/cufinufft.py
paquiteau: Good catch, let's fix this

chaithyagr (approved):
LGTM

PR #348, opened 2026-01-24
Add Toeplitz Embedding for Gram Operator

### LINKED ISSUE #320: Toeplitz Embedding
## Description & use cases

"Toeplitz NUFFT" are a way to efficiently compute the autoadjoint operator $A^HA$ without requiring NUFFT Operation. They can greatly speed up reconstruction and be used in optimization scheme, to avoid going back and forth in the k-space. Notably they are very usefull to solve the least-square problem with the normal equations: $A^HAx = A^H b$

## Suggested implementation
- have a `make_toeplitz` method for the operator, that create the Toeplitz embedding and return a callable (probably a dedicated class as we have state to preserve, it could also be a LinOps). 
- It should also work for off-resonance corrected operators.

References:
- J. A. Fessler, Sangwoo Lee, V. T. Olafsson, H. R. Shi and D. C. Noll, "Toeplitz-based iterative image reconstruction for MRI with correction for magnetic field inhomogeneity," in IEEE Transactions on Signal Processing, vol. 53, no. 9, pp. 3393-3402, Sept. 2005, doi: 10.1109/TSP.2005.853152

Surprisingly, we didn't have an issue tracking that.

paquiteau:
- Closes #320 
- Replaces #342 

- [x] Add a `toeplitz.py` module with function to compute and apply toeplitz kernel
- [x] Add documentation and minimal examples for the interface
- [x] Add a generic `gram_op` method to `FourierOperatorBase` for minimal support
- [x] Add test to compare with naive implementation. 
- [ ] Backend specific implementations (e.g `cufinufft`, `gpunufft` to handle host-device transfers and smaps cachin). *might be addressed in future PRs*

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -157,6 +157,7 @@ html_theme_options = {
             "sg_launcher_links",
             "colab_link",
         ],
+        "explanations/**": ["page-toc"],
     },
     "header_links_before_dropdown": 4,
     "icon_links": [
--- a/docs/explanations/mrinufft_convention.rst
+++ b/docs/explanations/mrinufft_convention.rst
@@ -12,21 +12,21 @@ The NUFFT Operator
 
 MRI-NUFFT provides a common interface for computing NUFFT, regardless of the chosen computation backend. All the backends implement the following methods and attributes. More technical details are available in the API reference.
 
-All MRI-NUFFT operators inherit from :class:`FourierOperatorBase` . The minimum signature of an MRI nufft operator is:
+All MRI-NUFFT operators inherit from :class:`mrinufft.operators.base.FourierOperatorBase` . The minimum signature of an MRI nufft operator is:
 
 .. code-block:: python
 
     class MRIOperator(FourierOperatorBase):
 
-          backend = "my-nufft-backend"
+        backend = "my-nufft-backend"
 
-          def __init__(samples: np.ndarray,
+        def __init__(samples: np.ndarray,
             shape:tuple[int,...],
             density: str | callable | np.ndarray | dict = False,
             n_coils: int = 1,
             n_batchs: int = 1,
             smaps: str | callable | np.ndarray | dict = False,
-            ):
+        ):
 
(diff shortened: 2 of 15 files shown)


chaithyagr:
I will proceed to review this, we can do the last one in a different issue!

NCastro-FAU:
Looking forward to using this one. 
Will this already work for 2D Spirals ?

I was looking at the code and I dont quite understand why it is failing.
however I am looking forward to implement this one in my current setup. Might give it a go after ISMRM.

chaithyagr:
The current version should work, but it needs more extensive tests to go in. Please do post your logs and it can help understand issues which we probably didn't anticipate and we can have better tests in place

chaithyagr:
It doesn't depend on the trajectory

paquiteau:
@NCastro-FAU I am interested in your feedback as well. Probably some better results can be obtained by tuning the hyperparameters of the NUFFT backend, (notably the `kernel_width` / `eps` and the oversampling factor for the grid). 

PS: I also did a rebase on top master so that we get the latest and greatest.

chaithyagr:
Before losing this, lets rebase and merge while ensuring all tests pass. I think again its better to have something rather than wait for the most perfect implementation (which we can get to while tracking in an issue). Its better than letting your current efforts getting stale.

paquiteau:
I also think it is in good shape, but there was some accuracy issues (densely sampled radial iirc). Probably some kind of density comp would help in this case.

chaithyagr:
Can you handle the tests? maybe we should slightly weaken the tests when toeplitz is on? Also, gpuNUFFT does support toeplitz too, if you want to use it (on latest version). It isnt that much faster though..

paquiteau:
@chaithyagr Finally, all green ! (The gpunufft bugs is some stream stuff, I will probably have a deeper look at it in a coming PR, toeplitz on GPU should basically be identical in term of code for gpunufft and cufinufft, since there is no more nufft plan call involved once the toeplitz kernel has been computed.

PR #350, opened 2026-02-04
Implement memory cleanup in smaps.py

chaithyagr:
Rather self explainatory


paquiteau (changes requested):

src/mrinufft/extras/smaps.py
paquiteau: I think you can turn these into lambda function below, to make the linter happy (and it will be nicer as well)

PR #352, opened 2026-02-09
Add back support for UTE slew ramps

chaithyagr:
We are getting back slew ramps which was closed earlier #330

### DIFF
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -25,8 +25,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          # We use 'opencode run' to feed the issue body to the agent.
-          # The 'build' agent has the ability to write files and use git.
           opencode run "Fix this issue: ${{ github.event.issue.title }}. Context: ${{ github.event.issue.body }}. 
-          Create a new branch, commit the fix, and open a pull request." --agent build
+          Create a new branch, commit the fix, and open a pull request. DO NOT EXIT until the entire task is completed" --agent build
 
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -21,6 +21,7 @@ jobs:
           persist-credentials: true 
 
       - name: Run OpenCode Fix
+        timeout-minutes: 60  # Explicitly allow up to 1 hour
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -27,7 +27,7 @@ jobs:
 
         run: |
           opencode run "TASK: Fix the issue: ${{ github.event.issue.title }}.
-          
+          CONTEXT:  ${{ github.event.issue.body }} 
           STRICT WORKFLOW:
           0. Understand what needs to be done
           1. Create branch: feat/fix-${{ github.event.issue.number }}
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -1,0 +1,31 @@
+name: OpenCode Auto-Fix
+on:
+  issues:
+    types: [labeled]
+
+jobs:
+  opencode-agent:
+    if: github.event.label.name == 'fix-it'
+    runs-on: self-hosted # Crucial: tells GitHub to use your local machine
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          # Allow the runner to push changes back
+          persist-credentials: true 
+
+      - name: Run OpenCode Fix
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # We use 'opencode run' to feed the issue body to the agent.
+          # The 'build' agent has the ability to write files and use git.
+          opencode run "Fix this issue: ${{ github.event.issue.title }}. Context: ${{ github.event.issue.body }}. 
+          Create a new branch, commit the fix, and open a pull request." --agent build
+
--- a/.github/workflows/opencode.yml
+++ b/.github/workflows/opencode.yml
@@ -24,7 +24,18 @@ jobs:
         timeout-minutes: 60  # Explicitly allow up to 1 hour
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
         run: |
-          opencode run "Fix this issue: ${{ github.event.issue.title }}. Context: ${{ github.event.issue.body }}. 
-          Create a new branch, commit the fix, and open a pull request. DO NOT EXIT until the entire task is completed" --agent build
+          opencode run "TASK: Fix the issue: ${{ github.event.issue.title }}.
+          
+          STRICT WORKFLOW:
+          1. Create branch: feat/fix-${{ github.event.issue.number }}
+          2. Apply code changes.
+          3. Stage changes using 'git add'.
+          4. Commit changes using 'git commit -m \"fix: ${{ github.event.issue.title }}\"'.
+          5. Push to origin.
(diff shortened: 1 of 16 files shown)


paquiteau (commented):
Some high-level discussion before diving into the code: 

- How does this compare to [`patch_center_anomaly`](https://mind-inria.github.io/mri-nufft/generated/_autosummary/mrinufft.trajectories.gradients.patch_center_anomaly.html#mrinufft.trajectories.gradients.patch_center_anomaly) and what should be done with this function then ? 
- Examples please :) 
- Tests could also be added, but that is less important (we mostly test trajectories with examples anyways)

- I am not a huge fan of the decorator proposed here in its current form. There should be at least some documentation added to the function (as we did for the `@with_numpy/cupy/torch` decorators). 

- Extra care should be taken when modifying the input arguments of the trajectories (changing `Ns` for instance will have a broader impact with repercussion outside the center, changing the bandwith of acquisition and the density of points).

src/mrinufft/trajectories/tools.py
```diff
@@ -1139,3 +1146,146 @@ def get_packing_spacing_positions(
...
+    -------
+    Callable
+        A decorator that modifies the trajectory function to include
+        slew rate ramps.
+
+    Notes
+    -----
+    - The decorator modifies the trajectory function to ensure that the
+        gradients at the specified `ramp_to_index` are adjusted to comply with
+        the maximum slew rate and gradient amplitude constraints.
```
paquiteau: repetition of the description of the function

src/mrinufft/trajectories/tools.py
```diff
@@ -1139,3 +1146,146 @@ def get_packing_spacing_positions(
...
+    method: str = "lp-minslew",
+    acq: Acquisition | None = None,
+) -> NDArray:
+    traj = func(**func_kwargs)
+    acq = acq or Acquisition.default
+    unnormalized_traj = unnormalize_trajectory(traj, acq=acq)
+    gradients, initial_positions = convert_trajectory_to_gradients(traj, acq=acq)
+    gradients_to_reach = gradients[:, ramp_to_index]
+    # Calculate the number of time steps for ramps
+    min_length = min_length_connection(
```
paquiteau: some gracefull handling of errors is probably required here, i.e advise the user to use a larger ramp_to_index. Or maybe we loop until we go to far away (for instance 10% of KMAX ? )

src/mrinufft/trajectories/tools.py
```diff
@@ -1139,3 +1146,146 @@ def get_packing_spacing_positions(
...
+
+def _add_slew_ramp_to_traj_func(
+    func: Callable,
+    func_kwargs: dict,
+    ramp_to_index: int,
+    method: str = "lp-minslew",
+    acq: Acquisition | None = None,
+) -> NDArray:
+    traj = func(**func_kwargs)
+    acq = acq or Acquisition.default
```
paquiteau: I think some check should be done to see if the trajectory is already hardware compliant. If so, no need for all the following.

src/mrinufft/trajectories/tools.py
```diff
@@ -1139,3 +1146,146 @@ def get_packing_spacing_positions(
...
+    )
+    # Update the Ns of the trajectory to ensure we still give
+    # same Ns as users expect. We use extra 2 points as buffer.
+    n_slew_ramp = min_length + 2
+    func_kwargs["Ns"] -= n_slew_ramp - ramp_to_index
+    new_traj = func(**func_kwargs)
+    # Re-calculate the gradients
+    unnormalized_traj = unnormalize_trajectory(new_traj, acq=acq)
+    gradients, initial_positions = convert_trajectory_to_gradients(new_traj, acq=acq)
+    gradients_to_reach = gradients[:, ramp_to_index]
```
paquiteau: Careful, those are not the same condition as before for estimating the number of points. 

I don't like that it recomputes a new trajectory here, this could bypass the deal made with the user.

paquiteau (changes requested):

chaithyagr:
Can we merge this soon @paquiteau ? While i do think we disagree on things and I surely agree there are corner cases which we potentially mjiss currently, I believe this is better than nothing which is the current state. WDYT? If you agree, I will rebase changes and then this could be merged.

.github/workflows/opencode.yml
paquiteau, line 1: I don't think this belong here, we can discuss agent integration in another thread/issue/pr

src/mrinufft/operators/autodiff.py
paquiteau, line 1: This should not be in this pr as well, we are already handling this in #370

src/mrinufft/operators/base.py
paquiteau, line 1: same, #370 is the place for this

src/mrinufft/trajectories/tools/slew_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+        acq=acq,
+        N=n_slew_ramp,
+        method=method,
+    )[:, :-1]
+    ramp_up_traj = convert_gradients_to_trajectory(
+        gradients=ramp_up_gradients,
+        initial_positions=initial_positions,
+        acq=acq,
+    )
+    return np.hstack([ramp_up_traj, new_traj[:, ramp_to_index:]])
```
paquiteau, line 63: I think it would make more sense to first stitch the gradients waveforms, and then integrate to get the trajectory, this is closer to the physics, and probably saves you from offset errors.

src/mrinufft/trajectories/tools/slew_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def add_slew_ramp(
+    func: Callable | None = None,
+    ramp_to_index: int = 5,
+    acq: Acquisition | None = None,
+    slew_ramp_disable: bool = False,
+    method: str = "lp-minslew",
```
paquiteau, line 71: `"auto"` would be a better default here. (it will use `osqp` if available, and `lp-minslew` otherwise)

src/mrinufft/trajectories/tools/slew_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+        acq=acq,
+    )
+    return np.hstack([ramp_up_traj, new_traj[:, ramp_to_index:]])
+
+
+def add_slew_ramp(
+    func: Callable | None = None,
+    ramp_to_index: int = 5,
+    acq: Acquisition | None = None,
+    slew_ramp_disable: bool = False,
```
paquiteau, line 70: invert the inverted logic, just use `slew_ramp = True`

src/mrinufft/trajectories/tools/slew_ramp.py
paquiteau, line 71: also `method` is quite a generic argument name, that could clash with existing args. I would recommend making it a "private" argument, like `_slew_ramp_method`.

src/mrinufft/trajectories/tools/slew_ramp.py
```diff
@@ -0,0 +1,147 @@
...
+        gradients at the specified `ramp_to_index` are adjusted to comply with
+        the maximum slew rate and gradient amplitude constraints.
+    - If `slew_ramp_disable` is set to True, the trajectory function will
+        return the trajectory as is, without applying any slew ramps.
+    """
+
+    def decorator(trajectory_func):
+        sig = inspect.signature(trajectory_func)
+
+        @wraps(trajectory_func)
```
paquiteau, line 119: Since we modify the signature of the function, the docstring of the function should also be modified. We have done similar thing with the `with_<array>` decorator.

paquiteau (changes requested):
Thanks for reviving this ! 

First, There is some cleanup to do to avoid conflicts down the road. 

I do agree that slew ramp are an issue, and the proposed method is in fact a good solution, but I don't know if we should enforce it directly via the proposed decorator approach. 

In the API we already have `patch_center_anomaly` which does some similar thing, by working like a "manual" decorator. This might be a lighter /simpler approach here. 

Therefore I feel we could take a similar approach here, and add extensive documentation: in the docstring themselves, via a lightweight docstring decorator, and more broadly in #372.

PR #361, opened 2026-02-24
Add submodules to trajectories

Daval-G:
Just re-organizing the trajectory module to split the huge files that are `trajectory2D.py`, `trajectory3D.py`, `tools.py` and making sense of `display.py` & `display3D.py`. We already had started passively putting new trajectories aside but it was time to do it for old ones too.

Unit tests are coming but I thought it would be simpler to first move stuff around without changing anything else


paquiteau:
Well this is something I was always a bit afraid to do :D Thanks for this ! there is a missing import in the example (see the CI) otherwise. Let's also trigger a build doc to check how the api documentation behaves with this structure.

paquiteau:
✅ Documentation built! (examples were disabled)
[download the artifact](https://github.com/mind-inria/mri-nufft/actions/runs/22387522589/artifacts/5649546655)

paquiteau (changes requested):
`example_3d_trajectory_display` is failing on missing import

chaithyagr (approved):
LGTM

PR #364, opened 2026-03-16
Add support for viewed_as_real

chaithyagr:
A simple extra interfact to help working easily with deepinv without needing complex denoisers.

PR #370, opened 2026-05-12
Add TV reconstruction with DeepInverse PDCP to model-based MRI example

### LINKED ISSUE #382: Add example of TV reconstruction

MelissaBoutlendj:
This PR extends the model-based MRI reconstruction example with a Total Variation (TV) reconstruction baseline using the official DeepInverse PDCP optimizer.

The implementation compares:

* adjoint reconstruction,
* wavelet reconstruction with FISTA,
* TV reconstruction with PDCP.

Main changes:

* added a TV-regularized reconstruction baseline using DeepInverse PDCP;
* added support for `viewed_as_real=True` complex MRI tensors;
* adapted `TVPrior` to real-view complex tensors by packing the real/imaginary dimension into the channel dimension before applying TV regularization;
* added PSNR and SSIM evaluation metrics;
* added visualization comparing Adjoint, Wavelet-FISTA, and TV-PDCP reconstructions.

The PDCP formulation uses:

* the MRI forward operator as `K`,
* `L2Distance` as the data-fidelity term in k-space,
*a custom cost function to compute the TV reconstruction loss

## Tested
- `git diff --check`
- `python -m examples.extras.example_recon`
- `python -m sphinx.cmd.build docs docs_build -D plot_gallery=0`

### DIFF
--- a/examples/extras/example_recon.py
+++ b/examples/extras/example_recon.py
@@ -135,12 +135,12 @@ early_stop = True
 # %%
 # Instantiate the algorithm class to solve the problem.
 wavelet_model = optim_builder(
-   iteration="FISTA",
-   prior=wavelet,
-   data_fidelity=data_fidelity,
-   early_stop=early_stop,
-   max_iter=max_iter,
-   params_algo=params_algo,
+    iteration="FISTA",
+    prior=wavelet,
+    data_fidelity=data_fidelity,
+    early_stop=early_stop,
+    max_iter=max_iter,
+    params_algo=params_algo,
 )
 x_wavelet = wavelet_model(y, physics_complex)
 
@@ -210,6 +210,7 @@ tv = TVPrior(n_it_max=20)
 #
 # Both arguments of the data-fidelity term are k-space tensors..
 
+
 def pdcp_cost_fn(x, data_fidelity, prior, cur_params, y, physics):
     return data_fidelity(cur_params["K"](x), y) + cur_params["lambda"] * prior(x)
 
@@ -226,7 +227,6 @@ pdcp_model = PDCP(
     g_first=False,
     cost_fn=pdcp_cost_fn,
 )
-
 
 
 x_pdcp_real = pdcp_model(y_real, physics)
--- a/examples/extras/example_recon.py
+++ b/examples/extras/example_recon.py
@@ -61,12 +61,24 @@ y += noise_level * (torch.randn_like(y) + 1j * torch.randn_like(y))
 
 # %%
 # Setup the physics and prior
-# With viewed_as_real=True, complex tensors are represented as
-# real-valued tensors with a final dimension of size 2:
-# (..., 2) = (real part, imaginary part).
+# With ``viewed_as_real=True``, the MRI-NUFFT DeepInverse interface exposes
+# complex-valued MRI tensors through real-valued tensor representations
+# compatible with DeepInverse optimization methods.
+#
+# The wrapper automatically converts between:
+#
+# - image-space tensors:
+#   ``(B, 2C, D, H, W)`` real-packed
+#   ↔
+#   ``(B, C, D, H, W)`` complex
+#
+# - k-space tensors:
+#   ``(B, C, N, 2)``
+#   ↔
+#   ``(B, C, N)`` complex
 physics = fourier_op.make_deepinv_phy(viewed_as_real=True)
 
-# Complex-valued physics used for methods that operate directly on complex tensors.
+# Complex-valued physics for methods that need complex tensors directly
 physics_complex = fourier_op.make_deepinv_phy()
 
 y_real = torch.view_as_real(y.contiguous())
@@ -123,12 +135,12 @@ early_stop = True
 # %%
 # Instantiate the algorithm class to solve the problem.
 wavelet_model = optim_builder(
-    iteration="FISTA",
-    prior=wavelet,
-    data_fidelity=data_fidelity,
-    early_stop=early_stop,
-    max_iter=max_iter,
-    params_algo=params_algo,
+   iteration="FISTA",
+   prior=wavelet,
+   data_fidelity=data_fidelity,
+   early_stop=early_stop,
+   max_iter=max_iter,
+   params_algo=params_algo,
 )
 x_wavelet = wavelet_model(y, physics_complex)
 
@@ -138,9 +150,9 @@ x_wavelet = wavelet_model(y, physics_complex)
 # ----------------------------------------------------
 #
 # As an additional model-based reconstruction baseline, we reconstruct the
-# image using a Total Variation (TV) prior. TV regularization promotes images
-# that are piecewise smooth while preserving sharp edges, which is useful in
-# MRI where anatomical structures often have smooth regions separated by
+# image using a Total Variation (TV) prior. TV regularization promotes
+# piecewise-smooth images while preserving sharp edges, which is useful in MRI
+# because anatomical structures often contain smooth regions separated by clear
 # boundaries.
 #
 # We solve the following variational problem:
@@ -152,89 +164,60 @@ x_wavelet = wavelet_model(y, physics_complex)
 # where A is the non-Cartesian MRI forward operator, y is the measured k-space
 # data, and x is the reconstructed image.
 #
-# TVPrior is designed for real-valued tensors. With the new
-# `viewed_as_real=True` option, complex MRI data are represented as
-# real-valued tensors with a final dimension of size 2:
-# (..., 2) = (real part, imaginary part).
-# For 3D MRI data, this representation produces 6D tensors of shape
-# (B, C, D, H, W, 2), while DeepInverse TVPrior expects 4D or 5D inputs.
-# We therefore pack the final real/imaginary dimension into the channel
-# dimension before applying TVPrior, and unpack it afterwards.
-# This avoids manually splitting complex tensors into separate real and
-# imaginary TV proximal operations.
-# The TV regularization weight was selected separately using Optuna.
+# ``TVPrior`` operates on real-valued image tensors. With
+# ``viewed_as_real=True``, complex MRI images are exposed through a real-valued
+# channel representation:
+#
+# .. math::
+#
+#    (B, C, D, H, W)_{\mathbb{C}}
+#    \longrightarrow
+#    (B, 2C, D, H, W)_{\mathbb{R}}
+#
+# This keeps the tensor compatible with DeepInverse priors expecting 4D or 5D
+# real-valued image tensors, while internally preserving the complex-valued MRI
+# representation.
 
-lamb_tv = 0.05789015101052105
-
-
-class RealViewTVPrior(TVPrior):
-    """Adapt TVPrior to tensors represented with torch.view_as_real.
-
-    This helper packs the final real/imaginary dimension into the channel
-    dimension before applying TV regularization, then restores the original
-    layout afterwards.
-    """
-
-    def _pack(self, x):
-        if x.shape[-1] != 2:
-            return x, None
-
-        original_shape = x.shape
-        b, c, *spatial, two = original_shape
-        x = x.movedim(-1, 2)          # (B, C, 2, D, H, W)
-        x = x.reshape(b, c * 2, *spatial)  # (B, 2*C, D, H, W)
-        return x, original_shape
-
-    def _unpack(self, x, original_shape):
-        if original_shape is None:
-            return x
-
-        b, c, *spatial, two = original_shape
-        x = x.reshape(b, c, 2, *spatial)
-        x = x.movedim(2, -1)          # back to (B, C, D, H, W, 2)
-        return x
-
-    def prox(self, x, *args, gamma=None, **kwargs):
-        x_packed, original_shape = self._pack(x)
-        out = super().prox(x_packed, *args, gamma=gamma, **kwargs)
-        return self._unpack(out, original_shape)
-
-    def forward(self, x, *args, **kwargs):
-        x_packed, _ = self._pack(x)
-        return super().forward(x_packed, *args, **kwargs)
-
-
-tv = RealViewTVPrior(n_it_max=20)
+lamb_tv = 50
+tv = TVPrior(n_it_max=20)
 
 
 # %%
-# We now solve the same TV-regularized MRI reconstruction problem using the
+# We now solve the TV-regularized MRI reconstruction problem using the
 # official DeepInverse PDCP optimizer. PDCP implements a Chambolle-Pock
 # primal-dual splitting method for objectives of the form F(Kx) + lambda G(x).
 #
-# In the PDCP formulation, the problem is written as F(Kx) + lambda G(x).
-# Here, K is the MRI forward operator A, so F acts directly on k-space data.
-# We use L2Distance to compare A(x) and y, and define a custom cost function
-# so that the monitored objective is evaluated as L2Distance(A(x), y) + lambda TV(x).
+# In this formulation, K is the MRI forward operator A. Therefore, F acts on
+# the predicted k-space data A(x), not directly on the image x.
+#
+# We use ``L2Distance`` as the data-fidelity term. Since ``L2Distance`` directly
+# compares its two inputs, the monitored objective must be evaluated as:
+#
+# .. math::
+#
+#    \frac{1}{2}\|A(x) - y\|_2^2 + \lambda \operatorname{TV}(x)
+#
+# We therefore define a custom cost function to compare ``A(x)`` with the
+# measured k-space data ``y``. Without this custom cost function, the default
+# objective would incorrectly compare the image tensor ``x`` with the k-space
+# measurements ``y``.
+#
 # The optimization problem remains:
 #
 # .. math::
 #
 #    \min_x \frac{1}{2}\|Ax - y\|_2^2 + \lambda \operatorname{TV}(x)
 #
-# Since F acts directly on K(x) = A(x), both inputs of the data-fidelity
-# term are k-space tensors.
-
-data_fidelity_distance = L2Distance()
-
+# Both arguments of the data-fidelity term are k-space tensors..
 
 def pdcp_cost_fn(x, data_fidelity, prior, cur_params, y, physics):
     return data_fidelity(cur_params["K"](x), y) + cur_params["lambda"] * prior(x)
 
+
 pdcp_model = PDCP(
     K=physics.A,
     K_adjoint=physics.A_adjoint,
-    data_fidelity=data_fidelity_distance,
+    data_fidelity=L2Distance(),
     prior=tv,
     lambda_reg=lamb_tv,
     stepsize=stepsize,
@@ -244,8 +227,17 @@ pdcp_model = PDCP(
     cost_fn=pdcp_cost_fn,
 )
 
+
+
 x_pdcp_real = pdcp_model(y_real, physics)
-x_pdcp = torch.view_as_complex(x_pdcp_real.contiguous())
+
+
+# Convert from 5D real-packed to complex for visualization
+b, c2, *spatial = x_pdcp_real.shape
+c = c2 // 2
+x_pdcp = torch.view_as_complex(
+    x_pdcp_real.reshape(b, c, 2, *spatial).movedim(2, -1).contiguous()
+)
 
 
 # %%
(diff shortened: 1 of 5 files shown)


src/mrinufft/operators/autodiff.py
```diff
@@ -463,27 +463,51 @@ def _check_input_shape(
         return True
 
 
+def complex_view_wrapper(method):
+    """Handle real-view tensors for complex MRI operators."""
+    def wrapper(self, x: torch.Tensor, **kwargs):
+        if self.viewed_as_real:
+            if x.shape[-1] != 2:
```
paquiteau: I think this check is redundant with what torch.view_as_complex does already

https://docs.pytorch.org/docs/2.11/generated/torch.view_as_complex.html

https://github.com/pytorch/pytorch/blob/5bd9e674fb4de5d6bb00a8e185c50eb7c7b6b913/torch/_refs/__init__.py#L1112

src/mrinufft/operators/base.py
```diff
@@ -421,8 +421,10 @@ def make_deepinv_phy(self, *args, **kwargs) -> DeepInvPhyNufft:
 
         from mrinufft.operators.autodiff import DeepInvPhyNufft
 
+        viewed_as_real = kwargs.pop("viewed_as_real", False)
```
paquiteau, line 429: needs the corresponding docstring above as well

tests/operators/test_autodiff.py
```diff
@@ -372,3 +372,31 @@ def test_forward_and_grad(operator, interface):
...
+        smaps = batched_smaps_from_op(operator)
+        y_real = physics.A(img_real, smaps=smaps)
+        adj_real = physics.A_adjoint(y_real, smaps=smaps)
+    else:
+        y_real = physics.A(img_real)
+        adj_real = physics.A_adjoint(y_real)
+
+    assert img_real.shape[-1] == 2
+    assert y_real.shape[-1] == 2
+    assert adj_real.shape[-1] == 2
```
paquiteau: you can to more restrictive checks by comparing with the whole shape.

paquiteau (changes requested):
of to a good start!  

here are some initial comments, there are also some issues in linter and tests to be fixed.

MelissaBoutlendj:
Thanks for the review. I addressed the requested changes in the latest commit.
I also reran the targeted test and docs build locally.

src/mrinufft/operators/autodiff.py
MelissaBoutlendj: Fixed in 63222f8 by removing the redundant explicit check.

src/mrinufft/operators/base.py
MelissaBoutlendj: documented viewed_as_real in make_deepinv_phy;

tests/operators/test_autodiff.py
MelissaBoutlendj: Updated the test to compare the full tensor shapes in 63222f8.

examples/extras/example_recon.py
```diff
@@ -3,25 +3,39 @@
...
 from brainweb_dl import get_mri
-from deepinv.optim.prior import WaveletPrior
 from deepinv.optim.data_fidelity import L2
 from deepinv.optim.optimizers import optim_builder
-
+from deepinv.optim.prior import WaveletPrior, TVPrior
 from mrinufft import get_operator
 from mrinufft.trajectories import initialize_3D_cones
+import torch.nn as nn
+from deepinv.models import TVDenoiser
```
chaithyagr: DO you really need TV Prior and TV Denoiser?

examples/extras/example_recon.py
```diff
@@ -51,7 +65,15 @@
 
 # %%
 # Setup the physics and prior
-physics = fourier_op.make_deepinv_phy()
+physics_complex = fourier_op.make_deepinv_phy()
+
+# With viewed_as_real=True, complex tensors are represented as
+# real-valued tensors with a final dimension of size 2:
+# (..., 2) = (real part, imaginary part).
+physics_real = fourier_op.make_deepinv_phy(viewed_as_real=True)
```
chaithyagr:
```suggestion
physics = fourier_op.make_deepinv_phy(viewed_as_real=True)
```

I dont like calling it physics real

examples/extras/example_recon.py
```diff
@@ -61,49 +83,346 @@
...
 # %%
 # Setup and run the reconstruction algorithm
 # Data fidelity term
 data_fidelity = L2()
 # Algorithm parameters
 lamb = 1e1
-stepsize = 0.8 * float(1 / fourier_op.get_lipschitz_cst().get())
+L = fourier_op.get_lipschitz_cst()
+if hasattr(L, "get"):
+    L = L.get()
```
chaithyagr: Why complicate? 
Whatever earlier was good right? (Assuming GPU)

examples/extras/example_recon.py
```diff
@@ -61,49 +83,346 @@
...
-x_wavelet = wavelet_recon(y, physics)
+x_wavelet = wavelet_model(y, physics_complex)
+
+
+# %%
+# Total variation reconstruction with proximal gradient descent
+# ------------------------------------------------------------
+#
+# As an additional model-based reconstruction baseline, we reconstruct the
+# image using a Total Variation (TV) prior. TV regularization promotes images
```
chaithyagr: Personally I will only keep TVPrior with PDCP and not with PGD...

examples/extras/example_recon.py
```diff
@@ -61,49 +83,346 @@
...
+            p = p_old + self.rho * (p - p_old)
+
+            rel_err = torch.linalg.norm(
+                x_old.flatten() - x.flatten()
+            ) / (torch.linalg.norm(x.flatten()) + 1e-12)
+
+            if rel_err < self.stopping_criterion:
+                break
+
+        return x
```
chaithyagr: I would really avoid writing this. Do we really need this? Cant we just use `TVPrior` directly?

tests/operators/test_autodiff.py
```diff
@@ -372,3 +372,31 @@ def test_forward_and_grad(operator, interface):
...
+        smaps = batched_smaps_from_op(operator)
+        y_real = physics.A(img_real, smaps=smaps)
+        adj_real = physics.A_adjoint(y_real, smaps=smaps)
+    else:
+        y_real = physics.A(img_real)
+        adj_real = physics.A_adjoint(y_real)
+
+    assert img_real.shape == (*img_data.shape, 2)
+    assert y_real.shape == (*operator.op(img_data).shape, 2)
+    assert adj_real.shape == img_real.shape
```
chaithyagr: Good job on adding the test!

chaithyagr (changes requested):
Thank you so much for the PR! Great work. Please find my comments below.

MelissaBoutlendj:
@chaithyagr Thank you for the review. I have addressed the requested changes.

examples/extras/example_recon.py
```diff
@@ -51,7 +61,28 @@
...
+# - image-space tensors:
+#   ``(B, 2C, D, H, W)`` real-packed
+#   ↔
+#   ``(B, C, D, H, W)`` complex
+#
+# - k-space tensors:
+#   ``(B, C, N, 2)``
+#   ↔
+#   ``(B, C, N)`` complex
+physics = fourier_op.make_deepinv_phy(viewed_as_real=True)
```
paquiteau: I feel this explanation should be put down, when you present the TV algorithm

examples/extras/example_recon.py
```diff
@@ -3,23 +3,32 @@
...
+a 3D cones trajectory. We then compare several reconstruction approaches:
 
-# %%
+1. Adjoint reconstruction, providing a fast baseline with sampling artifacts.
+2. Wavelet-regularized reconstruction solved with FISTA.
+3. Total Variation reconstruction solved with the DeepInverse PDCP optimizer.
+
+The goal of this example is to illustrate how MRI-NUFFT physics operators can
+be coupled with DeepInverse optimization tools to solve model-based MRI inverse
+problems and compare different regularization priors.
```
paquiteau:
```suggestion
The goal of this example is to illustrate how MRI-NUFFT physics operators can
be coupled with DeepInverse optimization tools to solve model-based MRI inverse
problems and compare different regularization priors, suppporting both complex-valued and real-valued (applied on the real and imaginary part) regularizations.
```

examples/extras/example_recon.py
```diff
@@ -61,49 +92,208 @@
...
+# .. math::
+#
+#    (B, C, D, H, W)_{\mathbb{C}}
+#    \longrightarrow
+#    (B, 2C, D, H, W)_{\mathbb{R}}
+#
+# This keeps the tensor compatible with DeepInverse priors expecting 4D or 5D
+# real-valued image tensors, while internally preserving the complex-valued MRI
+# representation.
+
```
paquiteau: insert the view_as_real operator here

src/mrinufft/operators/autodiff.py
```diff
@@ -463,27 +463,82 @@ def _check_input_shape(
...
 
 
+def complex_view_wrapper(method):
+    """Handle real-view tensors for complex MRI operators.
+
+    Supports both single and multiple tensor arguments.
+    Auto-detects tensor dimensionality and applies appropriate conversion.
+
+    - Image-space (input to forward op): (B, 2C, D, H, W) - 5D real-packed
+    - K-space (output of forward op): (B, C, N, 2) - 4D real-packed
```
paquiteau:
```suggestion
    - Image-space (input to forward op): (B, 2C, *XYZ) - 5D real-packed
    - K-space (output of forward op): (B, C, K, 2) - 4D real-packed
```

To be consistent with the conventions in the various interfaces in `operators` module.

src/mrinufft/operators/base.py
```diff
@@ -402,6 +402,11 @@ def make_deepinv_phy(self, *args, **kwargs) -> DeepInvPhyNufft:
             If provided, specifies batch size for varying data/smaps pairs.
             Default is None, which means no batching
 
+        viewed_as_real : bool, optional
+            If True, the DeepInverse physics wrapper accepts and returns
+            real-view tensors with a final dimension of size 2 representing
+            the real and imaginary parts. Default is False.
```
paquiteau, line 408: Also mention the channel-packed setup !

src/mrinufft/operators/autodiff.py
```diff
@@ -463,27 +463,82 @@ def _check_input_shape(
...
+        elif is_kspace:
+            b, c, *spatial = out.shape
+            out = torch.view_as_real(out)
+            out = out.movedim(-1, 2)
+            out = out.reshape(b, c * 2, *spatial)
+            return out
+
+        return out
+
+    return wrapper
```
paquiteau: There are several issues with this decorator: 
1. This does not support 2D image, and relies on the dimensions of the input tensor to know if you are in image space or not. 
2. A decorator / wrapper should do minimalistic stuff, and extract any complex modification to specific functions. 
3. I am not a fan of the `.contiguous()` method, as it will potentially create a new memory copy. Is it really necessary here ? 
4. Ideally, you should call @wraps from functools to preserve the signature and docstring

I propose you rewrite this and split the concerns; I will let you fill in the blanks; don't forget docstrings ;) 

```python
def kspace_as_real(y):
    """View k-space data as real data ... """
    ... 

def kspace_as_cpx(y): 
    """View k-space data as real data ... """
    ... 

def image_as_real(x): 
    """View image data as real channel-packed data ... """ 
    ...

def image_as_cpx(x): 
    ...

def complex_view_wrapper(in, out):
    in_view = kspace_as_cpx if in == "ksp" else image_as_cpx
    out_view = kspace_as_real if in == "ksp" else image_as_real
    def decorator(method):
        @wraps(method)
        def wrapper(self, x, *args, **kwargs):
            if not self.viewed_as_real: 
                return method(self, x, *args, **kwargs)
            return out_view(method(self, in_view(x), *args, **kwargs)
        return wrapper
    return decorator
```

examples/extras/example_recon.py
```diff
@@ -61,49 +92,208 @@
...
+# more informative for images because it compares local contrast and structure.
+#
+# Metrics are computed on magnitude images, since the reconstructions are
+# complex-valued.
+
+
+# %%
+from deepinv.loss.metric import PSNR, SSIM
+
+psnr = PSNR()
```
paquiteau: When running the example I get the following values: 

```
Adjoint PSNR: -62.08
Wavelet PSNR: -47.19
TV-PDCP PSNR: -44.63
Adjoint SSIM: 0.3945
Wavelet SSIM: 0.5691
TV-PDCP SSIM: 0.6284
```

This is bad, you should set `max_pixel=None` to have positive and (expected) values: https://deepinv.github.io/deepinv/api/stubs/deepinv.loss.metric.PSNR.html

paquiteau (commented):
Thanks for the update, There is a review-in-the-review regarding viewed_as_real, as it was taken from #364.

examples/extras/example_recon.py
```diff
@@ -61,49 +92,208 @@
...
+
+
+def pdcp_cost_fn(x, data_fidelity, prior, cur_params, y, physics):
+    return data_fidelity(cur_params["K"](x), y) + cur_params["lambda"] * prior(x)
+
+
+pdcp_model = PDCP(
+    K=physics.A,
+    K_adjoint=physics.A_adjoint,
+    data_fidelity=L2Distance(),
```
paquiteau, line 158: Should be L2(). L2distance is a distance object not a data_fidelity

paquiteau (changes requested):

examples/extras/example_recon.py
chaithyagr: I think we can just remove it, its doc for old code :)

examples/extras/example_recon.py
```diff
@@ -51,7 +61,28 @@
...
+#   ``(B, C, D, H, W)`` complex
+#
+# - k-space tensors:
+#   ``(B, C, N, 2)``
+#   ↔
+#   ``(B, C, N)`` complex
+physics = fourier_op.make_deepinv_phy(viewed_as_real=True)
+
+# Complex-valued physics for methods that need complex tensors directly
+physics_complex = fourier_op.make_deepinv_phy()
```
chaithyagr: So, turns out we can just use physics_complex even with wavelet (by setting `is_complex` to `False` which is default).
Maybe we can simplify the example with that @paquiteau ? 
Idea is we want to show that we can directly interface with any prior is deepinv

chaithyagr:
```suggestion
physics = fourier_op.make_deepinv_phy(viewed_as_real=True)
```

Lets call it just physics and not real or complex, esp if we use real version everywhere.,

examples/extras/example_recon.py
```diff
@@ -51,7 +61,28 @@
...
+
+y_real = torch.view_as_real(y.contiguous())
```
chaithyagr: We need to have this coversion to real written somewhere in `mri-nufft` to ensure users have access to it

Same goes for real to imaginary conversion. I will detail this below

examples/extras/example_recon.py
chaithyagr:
```suggestion
```

We remove this if we want it to wor with "real" version of NUFFT

examples/extras/example_recon.py
chaithyagr: Again, doc for old code, please cleanup here

examples/extras/example_recon.py
```diff
@@ -61,49 +92,208 @@
...
+#
+# .. math::
+#
+#    \min_x \frac{1}{2}\|Ax - y\|_2^2 + \lambda \operatorname{TV}(x)
+#
+# Both arguments of the data-fidelity term are k-space tensors..
+
+
+def pdcp_cost_fn(x, data_fidelity, prior, cur_params, y, physics):
+    return data_fidelity(cur_params["K"](x), y) + cur_params["lambda"] * prior(x)
```
chaithyagr, line 148: I really dont like this, we need to discuss with folks in deepinv to understand whats up

examples/extras/example_recon.py
chaithyagr: Yaaaaaa, about this, weirdly, the PDCP algorithm internally has data fidelity as:

data_fidelity( physics.op ( x ) , y)


While L2 implements 
physics.op(x) - y



If you give L2 here we end up doing


data_fidelity(physics.op(physics.op(x)), y)


Again, i find this bizzare and we need folks from deepinv tell us why they did this differently compared to PGD where its relatively more straightforward

examples/extras/example_recon.py
```diff
@@ -61,49 +92,208 @@
...
+
+x_pdcp_real = pdcp_model(y_real, physics)
+
+
+# Convert from 5D real-packed to complex for visualization
+b, c2, *spatial = x_pdcp_real.shape
+c = c2 // 2
+x_pdcp = torch.view_as_complex(
+    x_pdcp_real.reshape(b, c, 2, *spatial).movedim(2, -1).contiguous()
 )
```
chaithyagr: the util function here must be in mri-nufft

src/mrinufft/operators/autodiff.py
```diff
@@ -463,27 +463,82 @@ def _check_input_shape(
...
+        is_kspace = x.ndim == 4
+
+        # Handle image-space input conversion (B, 2C, D, H, W) → complex (B, C, D, H, W)
+        if is_image_space:
+            b, c2, *spatial = x.shape
+            c = c2 // 2
+            x = x.reshape(b, c, 2, *spatial)
+            x = x.movedim(2, -1)
+            # Convert to complex: (B, C, D, H, W, 2)
+            x = torch.view_as_complex(x.contiguous())
```
chaithyagr: Make this as a separate function and add to mrinufft.converters ? @paquiteau ideas for better name?

src/mrinufft/operators/autodiff.py
```diff
@@ -463,27 +463,82 @@ def _check_input_shape(
...
+        # Handle k-space input conversion (B, C, N, 2) → complex (B, C, N)
+        elif is_kspace:
+            # Directly convert real-packed k-space to complex tensor
+            x = torch.view_as_complex(x.contiguous())
```
chaithyagr: Same as above

src/mrinufft/operators/autodiff.py
chaithyagr: Lol, just saw this now, we need this, but with each exposed for user (maybe directly importable as 
`from mrinufft import kspace_as_real` ?

Or maybe under `compat`?

chaithyagr (changes requested):
Great work! we are almost gettiong there. a bunch of coding style changes

examples/extras/example_recon.py
paquiteau: Indeed, but you still need the view as real setup for the TV case anyway ?

paquiteau (approved):
Thank for the modifications, this is in good enough shape for me. 

However, there is no TV example anymore, is this on purpose ? In this case why not just rely on the `is_complex=True` of the wavelet Prior and save us all this trouble ? @chaithyagr

MelissaBoutlendj:
@paquiteau Thank you for the feedback! Yes, the TV example was intentionally removed for now because PDCP does not yet work with the current viewed_as_real wrapper.

chaithyagr:
@MelissaBoutlendj , i still feel you should add the current working setup of PDCP just to show how to use it anyway...

chaithyagr:
We can always fix it if we get response from folks from deepinv

MelissaBoutlendj:
@chaithyagr no worries I will added it in a try/except block

examples/extras/example_recon.py
```diff
@@ -43,67 +54,181 @@
...
+
+y_real = kspace_as_real(y).float()
+
 wavelet = WaveletPrior(
     wv="sym8",
     wvdim=3,
     level=3,
-    is_complex=True,
+    is_complex=False,
 )
```
paquiteau, line 75: I liked the previous version where there was also the show case of a complex-valued prior (so with `viewed_as real=False`, and `is_complex=True`). At least,  there should be some more explanation in the example on how / why set the values this or the other way.

chaithyagr: I think we can explain but in the end the idea is we can just straight off use any prior without much changes in code on deepinv side.

MelissaBoutlendj: Thank you for the feedback,  is it better to add both cases side by side in the same example?

Complex-valued prior : physics = fourier_op.make_deepinv_phy() with WaveletPrior(is_complex=True) using y directly
Real-valued prior :  physics = fourier_op.make_deepinv_phy(viewed_as_real=True) with WaveletPrior(is_complex=False) and TVPrior using y_real
 and also add more explanation.

chaithyagr (approved):
Thank you for all the changes. Looks good to me. We can merge when green with all the tests

paquiteau (approved):

PR #384, opened 2026-06-09
Add non-Cartesian self-supervised UNet example using SSDU

MelissaBoutlendj:
This PR adds a self-supervised MRI reconstruction example alongside  the existing supervised UNet example (`example_fastMRI_UNet.py`).

Both models are trained side by side on the same BrainWeb image and 
cones trajectory, allowing direct comparison between supervised and 
self-supervised reconstruction.

## Changes

  - Same architecture, trajectory, and image as the supervised example
  - K-space split into Theta (60%) and Lambda (40%) per cone
  - SSDU loss: compares predicted k-space at Lambda with true measurements
  - No ground truth needed during training

## References

- SSDU: Yaman et al., MRM 2020. 
- DDSS: Zhou et al., Medical Image Analysis 2022.


examples/extras/example_recon.py
```diff
@@ -12,13 +12,11 @@
 1. Adjoint reconstruction, providing a fast baseline with sampling artifacts.
```
philouc: density compensated adjoint NUFFT ?

examples/extras/example_recon.py
```diff
@@ -12,13 +12,11 @@
...
 2. Wavelet-regularized reconstruction solved with FISTA.
```
philouc: Synthesis prior wavelet based regularization

examples/extras/example_recon.py
```diff
@@ -12,13 +12,11 @@
...
 3. Total Variation reconstruction solved with the DeepInverse PDCP optimizer.
-.
+
 The goal of this example is to illustrate how MRI-NUFFT physics operators can
 be coupled with DeepInverse optimization tools to solve model-based MRI inverse
 problems and compare different regularization priors, supporting both
```
philouc: regularization terms (or priors)

examples/extras/example_recon.py
philouc, line 62: Please explain why 0.02%?

philouc, line 119: You cannot solve this ANALYSIS form unless W is an Orthonormal Wavelet transform
Please fix that and turn it into a SYNTHESIS form (as you're using FISTA)

philouc (commented):
Comments to be revised.

examples/extras/example_recon.py
paquiteau: here we use the `sym8` wavelet transform, so the analysis formulation is ok :)

examples/GPU/example_fastMRI_UNet.py
```diff
@@ -112,79 +118,194 @@ def plot_state(axs, mri_2D, traj, recon, loss=None, save_name=None):
...
-fig, axs = plt.subplots(2, 2, figsize=(10, 10))
-fig.suptitle("Training Starting")
-axs = axs.flatten()
+idx_theta, idx_lambda = [], []
+for s in range(N_CONES):
+    # Get the indices of the points in the current cone and randomly permute them.
+    cone_pts = np.arange(s * N_SAMPLES, (s + 1) * N_SAMPLES)
+    perm = rng.permutation(N_SAMPLES)
+    idx_theta.append(cone_pts[perm[:n_theta_per_cone]])
+    idx_lambda.append(cone_pts[perm[n_theta_per_cone:]])
```
paquiteau: Lets make this more explicit 
```suggestion
for s in range(N_CONES):
    # Get the indices of the points in the current cone and randomly permute them.
    locs = rng.choices(N_SAMPLES, n_theta_per_cones, replace=False)
    mask = np.zeros(N_SAMPLES,dtype=bool)
    mask[locs] = True
    idx_theta.append(cone_pts[mask])
    idx_lambda.append(cones_pts[~mask])
```

Another solution would be to use train_test_split from scikit-learn :)

Also, since we have boolean mask, it might be also used directly, instead of integer indexing

examples/GPU/example_fastMRI_UNet.py
```diff
@@ -112,79 +118,194 @@ def plot_state(axs, mri_2D, traj, recon, loss=None, save_name=None):
...
-    np.abs(dc_adjoint[0][0].detach().cpu().numpy()),
-    cmap="gray",
+
+# %%
+# NUFFT operator for Lambda — used only in the SSDU loss.
+op_lambda = get_operator(BACKEND, wrt_data=True)(
+    traj_lambda,
+    shape=(256, 256),
+    density=True,
+    squeeze_dims=False,
```
paquiteau, line 184: Using two operator can be memory intensive, to showcase mri-nufft capabilities it could also be possible to update the sample location on the fly (@chaithyagr what do you think ?)

paquiteau (commented):
Thanks for extending the example, here are a few remarks :)

examples/GPU/example_fastMRI_UNet.py
```diff
@@ -24,10 +24,11 @@
     {||\mathbf{y}_\Lambda||_1}
```
philouc, line 25: Somewhere in the doc of this class , you should say quite early that this example only works in 2D (slice by slice)

examples/GPU/example_fastMRI_UNet.py
```diff
@@ -37,210 +38,232 @@
...
 )
 
 
 # %%
-# Setup both models, same architecture, different input trajectories.
-device = "cuda" if torch.cuda.is_available() else "cpu"
+# SSDU loss predicts k-space at Lambda from the reconstruction and
+# compares it with the true measurements. No ground truth needed.
+def ssdu_loss(recon, kspace_lambda, op_lambda):
+    """Compute the SSDU self-supervised loss.
```
philouc, line 192: in the k-space domain

examples/GPU/example_fastMRI_UNet.py
```diff
@@ -287,56 +315,49 @@ def train():
...
 # References
 # ==========
 #
-# .. [fastmri] O. Ronneberger, P. Fischer, and Thomas Brox. U-net: Convolutional networks
-#           for biomedical image segmentation. In International Conference on Medical
-#           image computing and computer-assisted intervention, pages 234–241.
-#           Springer, 2015.
-#           https://github.com/facebookresearch/fastMRI/blob/main/fastmri/models/unet.py
+# .. [fastmri] O. Ronneberger, P. Fischer, and Thomas Brox. U-net:
+#           Convolutional networks for biomedical image segmentation.
```
philouc: If you work on brain fastMRI, the right reference is Muckley et al, IEEE TMI 2021 (which we cosigned with Zaccharie Ramzi and Jean-Luc Starck)

philouc (changes requested):

examples/GPU/example_fastMRI_UNet.py
MelissaBoutlendj: Thank you for the review

PR #386, opened 2026-06-18
Reorganizing imports

### LINKED ISSUE #327: Intersphinx not working fully
Tracking https://github.com/sphinx-gallery/sphinx-gallery/issues/1521

### LINKED ISSUE #383: Updating the public  API and imports
MRI-NUFFT has grown organically over the years, and as a result, the public facing API has become quite a mess, with imports available in multiple places and a lack of consistency in naming and organization 

This issue is to propose a (breaking) change to the public facing API and imports to make it more intuitive and easier for end-users

### Use absolute imports everywhere in the codebase
  This will make it clear where each function and class is coming from, and will
  help avoid issues with relative imports when the module structure changes.
  For example, instead of using `from .operators import get_operator`, we would
  use `from mrinufft.operators import get_operator` throughout the codebase.

### Limit the public API to a single level namespace, like in `scipy`. 

The current module structure is quite deep, so we flatten it up to the top-level modules, leaving the `mrinufft` namespace to core functionality. In this new structure the public API would be organized as follows:

  - `mrinufft`: `get_operator`, `Acquisition` and array utilities (`to_device`, `to_numpy`, etc.), and any other core functions that are widely used across the codebase
  - `mrinufft.operators`: any operator specific functions. However there are usually directly available as method once a nufft operator is created.
  - `mrinufft.trajectories`: `initialize_...` trajectories functions, projections, etc... 
  - `mrinufft.displays`: all functions related to visualizations and displays (extracted from the `mrinufft.trajectories` module)
  - `mrinufft.extras`, `.io`, `density`: same scope as before, but with a flat structure.

### Use lazy imports for operators and heavy dependencies 
Ynnecessary imports and reduce the initial load time of the library, notably because we test for the presence of heavy dependencies (like `torch`) by trying to import them, making overall import cycle slow (#7)

The solution is to use lazy imports for these modules, so that they are only imported when they are actually needed. This can likely be achieved using `importlib`  and some `__getattr__` magic in the `__init__.py` files of each module.

Another solution will be the `lazy import` (PEP 810), but will only be available in Python 3.15, we cannot rely on it for now, as we have to support Python 3.10 and 3.11 for the foreseeable future.

### Clean up the documentation API structure to reflect the new organization

The sphinx auto api is currently organized on a nested architecture, with auto documentation of any submodule, with the new namespaces, we can flatten it as well.

This will likely also hep for #327.
  

## Implementation Plan

- [ ] Use absolute imports throughout the codebase, and update all import statements accordingly. (`ruff` will help with this)
- [ ] Refactor the codebase to flatten the module structure and organize the public API as proposed.
- [ ] Implement lazy imports for operators and heavy dependencies.
- [ ] Update the documentation to reflect the new API structure and organization.
- [ ] Communicate the breaking changes to users and provide a migration guide if necessary.
- [ ] Update tests to reflect the new API structure and ensure that all tests pass with the new organization.
- [ ] Update examples and tutorials to reflect the new API structure and organization.

paquiteau:
- Closes #383
- Closes #327 

- [x] Completely rework the namespace of MRI-NUFFT, following #383
- [x] Update the docs build accordingly
- [x] Fix the imports in example : This will also repair intersphinx !
- [x] Almost nothing at top level (for now there is `get_operator` and the submodules, a bit like `scipy` namespace)

Not adressed here:
- Update the documentation of modules to be more than a single line


paquiteau:
❌ Failed to build documentation 
[Check Log output](https://github.com/mind-inria/mri-nufft/actions/runs/27790763430)

docs/conf.py
```diff
@@ -45,7 +45,7 @@
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
     "sphinx.ext.mathjax",
-    "sphinx.ext.linkcode",
+    #   "sphinx.ext.linkcode",
```
chaithyagr: why though? no more linkcode cause its broken?

chaithyagr (approved):
LGTM, the review here is hard, im assuming you followed what you said in the issue

docs/conf.py
paquiteau: it was a temporary fix, thanks for catching this

paquiteau:
I made the git commit history as clean as possible to help referencing further issues, it should be easier to refere to it later if needed.

PR #392, opened 2026-07-01
feat(cufinufft): use stream  for H2D2H computation, and tweaks.

### LINKED ISSUE #383: Updating the public  API and imports

paquiteau:
- Closes #341 
- On top of #348 

- Add support for stream in cufinufft backend for H2D2H computations. 
- Minor performance improvement with norm_factor (it was eating up a whole s when done on cpu with in-place divide !)

### DIFF
--- a/src/mrinufft/operators/interfaces/cufinufft.py
+++ b/src/mrinufft/operators/interfaces/cufinufft.py
@@ -21,6 +21,7 @@ from mrinufft._array_compat import (
     pin_memory,
     get_array_module,
     auto_cast,
+    _array_to_cupy,
 )
 from mrinufft.operators.interfaces.utils import nvtx_mark
 
@@ -305,6 +306,7 @@ class MRICufiNUFFT(FourierOperatorBase, _ToggleGradPlanMixin):
                 self.log.warning("n_coils updated via smaps.")
                 self.n_coils = C
             if self.smaps_cached or is_cuda_array(new_smaps):
+                new_smaps = _array_to_cupy(new_smaps)
                 self.smaps_cached = True
                 self.log.warning(
                     "%s used on gpu for smaps.",
--- a/src/mrinufft/operators/interfaces/cufinufft.py
+++ b/src/mrinufft/operators/interfaces/cufinufft.py
@@ -1,5 +1,6 @@
 """Provides Operator for MR Image processing on GPU."""
 
+import weakref
 import numpy as np
 from numpy.typing import NDArray
 from mrinufft.operators.base import (
@@ -28,6 +29,21 @@ try:
     import cupy as cp
     from cufinufft import Plan
     from cufinufft._compat import get_array_ptr
+
+    _coil_combine_kernel = cp.ElementwiseKernel(
+        "raw T data, raw T smaps, int64 b, int32 n_t, int64 vol",
+        "raw T img",
+        """
+        long long off = b * vol + i;
+        for (int t = 0; t < n_t; t++) {
+            T d = data[t * vol + i];
+            T s = smaps[t * vol + i];
+            img[off] += d * T(s.real(), -s.imag());
+        }
+        """,
+        "coil_combine_kernel",
+    )
+
 except ImportError:
     CUFINUFFT_AVAILABLE = False
 
@@ -167,6 +183,12 @@ class MRICufiNUFFT(FourierOperatorBase, _ToggleGradPlanMixin):
         If True, will try to remove the singleton dimension for batch and coils.
     n_trans: int, default 1
         Number of transform to perform in parallel by cufinufft.
+    async_transfer: bool, default False
+        If True, pipeline host<->device transfers with compute (double
+        buffering) in the host-input code paths, using dedicated
+        non-blocking CUDA streams. This overlaps H2D/D2H copies with
+        cufinufft compute across batches, at the cost of extra pinned
+        host buffers and device buffers (roughly 2x the per-batch memory).
     kwargs :
         Extra kwargs for the raw cufinufft operator
 
@@ -201,6 +223,7 @@ class MRICufiNUFFT(FourierOperatorBase, _ToggleGradPlanMixin):
         verbose=False,
         squeeze_dims=False,
         n_trans=1,
+        async_transfer=False,
         **kwargs,
     ):
         # run the availaility check here to get detailled output.
@@ -215,12 +238,11 @@ class MRICufiNUFFT(FourierOperatorBase, _ToggleGradPlanMixin):
         except Exception as e:
             self.log.warning("Failed to set CUDA device %s: %s", gpu_device_id, e)
         super().__init__()
-        if (n_batchs * n_coils) % n_trans != 0:
-            raise ValueError("n_batchs * n_coils should be a multiple of n_transf")
-
         self.shape = shape
         self.n_batchs = n_batchs
         self.n_trans = n_trans
+        self.async_transfer = async_transfer
+        self._host_registry = {}
         self.squeeze_dims = squeeze_dims
         self.n_coils = n_coils
         self.autograd_available = True
@@ -245,6 +267,18 @@ class MRICufiNUFFT(FourierOperatorBase, _ToggleGradPlanMixin):
             raise ValueError(
                 "Smaps should be either a C-ordered np.ndarray, or a GPUArray."
             )
+
+        # n_trans must tiles the number of coils and batchs
+        if (n_batchs * n_coils) % n_trans != 0:
+            raise ValueError(
+                f"n_trans={n_trans} must divide n_batchs*n_coils={n_batchs * n_coils}"
+            )
+        # for optimal sense, n_trans must divide n_coils.
+        if self.uses_sense:
+            if self.n_coils % n_trans != 0:
+                raise ValueError(
+                    f"n_trans={n_trans} must divide n_coils={self.n_coils}"
+                )
         self.raw_op = RawCufinufftPlan(
             self._samples,
             tuple(shape),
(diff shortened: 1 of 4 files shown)


src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -29,6 +30,21 @@
...
+        long long off = b * vol + i;
+        for (int t = 0; t < n_t; t++) {
+            T d = data[t * vol + i];
+            T s = smaps[t * vol + i];
+            img[off] += d * T(s.real(), -s.imag());
+        }
+        """,
+        "coil_combine_kernel",
+    )
+
```
chaithyagr, line 47: Umm, cant you just use vector multiply provided by cupy? Any advantage of writing CUDA codes here?

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -202,6 +224,7 @@ def __init__(
         verbose=False,
         squeeze_dims=False,
         n_trans=1,
+        async_transfer=False,
```
chaithyagr, line 227: If it is indeed faster, I recommend default as True

chaithyagr:
Im still reviewing, but it would be good if you can post the nsight screenshots here while using and without using your new feature to be sure its working.

paquiteau:
Sure !
# Results 
<details>
<summary> Forward </summary>
image

</details>
<details>
<summary> Adjoint </summary>

image
</details>

```
op   (sync) : 141.35 ms ± 0.15 ms
op   (async): 104.13 ms ± 0.06 ms
op   speedup: 1.36x

adj  (sync) : 244.58 ms ± 0.15 ms
adj  (async): 219.04 ms ± 0.07 ms
adj  speedup: 1.12x
```
## Implementation 

<details> <summary> The script </summary>

```python 
from time import perf_counter

import numpy as np

from mrinufft import get_operator
from mrinufft.trajectories import initialize_3D_cones

N_REPEATS = 3
N_WARMUP = 1

shape = (192, 192, 128)
n_coils = 16
n_batchs = 1
n_trans = 4

samples = initialize_3D_cones(48, 13120).astype(np.float32)

smaps = 1j * np.random.rand(n_coils, *shape) + np.random.rand(n_coils, *shape)
smaps = smaps.astype(np.complex64)
smaps /= np.linalg.norm(smaps, axis=0)

common_kwargs = dict(
    n_coils=n_coils,
    smaps=smaps,
    n_batchs=n_batchs,
    n_trans=n_trans,
    squeeze_dims=False,
    gpu_method=1,
)

op_sync = get_operator("cufinufft")(
    samples, shape, async_transfer=False, **common_kwargs
)
op_async = get_operator("cufinufft")(
    samples, shape, async_transfer=True, **common_kwargs
)

image_data = (1j * np.random.rand(*op_sync.img_full_shape)).astype(op_sync.cpx_dtype)
image_data += np.random.rand(*op_sync.img_full_shape).astype(op_sync.cpx_dtype)

kspace_data = (1j * np.random.rand(*op_sync.ksp_full_shape)).astype(op_sync.cpx_dtype)
kspace_data += np.random.rand(*op_sync.ksp_full_shape).astype(op_sync.cpx_dtype)


def _time(fn, *args, n_repeats=N_REPEATS, n_warmup=N_WARMUP):
    for _ in range(n_warmup):
        fn(*args)
    times = []
    for _ in range(n_repeats):
        tic = perf_counter()
        fn(*args)
        toc = perf_counter()
        times.append(toc - tic)
    return np.mean(times), np.std(times)


op_sync_mean, op_sync_std = _time(op_sync.op, image_data)
op_async_mean, op_async_std = _time(op_async.op, image_data)

adj_sync_mean, adj_sync_std = _time(op_sync.adj_op, kspace_data)
adj_async_mean, adj_async_std = _time(op_async.adj_op, kspace_data)

print(f"shape={shape}, n_coils={n_coils}, n_batchs={n_batchs}, n_trans={n_trans}\n")
print(f"op   (sync) : {op_sync_mean * 1e3:.2f} ms ± {op_sync_std * 1e3:.2f} ms")
print(f"op   (async): {op_async_mean * 1e3:.2f} ms ± {op_async_std * 1e3:.2f} ms")
print(f"op   speedup: {op_sync_mean / op_async_mean:.2f}x\n")

print(f"adj  (sync) : {adj_sync_mean * 1e3:.2f} ms ± {adj_sync_std * 1e3:.2f} ms")
print(f"adj  (async): {adj_async_mean * 1e3:.2f} ms ± {adj_async_std * 1e3:.2f} ms")
print(f"adj  speedup: {adj_sync_mean / adj_async_mean:.2f}x")
```

</details>

### Setup: 
- RTX Ada 5000, cuda 13, driver 580
- I am using lastest master of cufinufft (there are also discussing performance regression[^0])
- The baseline shown here is already better than mri-nufft master as it also benefited from the `/= norm_factor -> *= inv_norm`  tweak. Here are some quantitative result
- I set `gpu_method=1`, which also speed things up. (we had poor SM warp occupancy with the heuristic-driven gpu_method=0 (which decided to use gpu_method=2) [^1]


-> This call for a broader benchmark :upside_down_face: 

[^0]:  https://github.com/flatironinstitute/finufft/pull/873
[^1]: https://finufft.readthedocs.io/en/latest/c_gpu.html#algorithm-performance-options

paquiteau:
Re on this,  I have had more though on it: 
- smaps locations: I think we should not disban the use of smaps on GPU, there are cases where it make sense to let them on the GPU (typically when you want to learn them)
- Async as a default: For now it is only opt-in, I feel we should make it opt-out, or just plainly mandatory (I don't see any downside for it) This will reduce the code base (basically removing `_{adj_op,op}_{calibless,sense}_host` functions. 
- We can also do Async Gram Operator / data-consistency, @chaithyagr do you want this in this PR as well ?

chaithyagr:
we can indeed do all async stuff on the same PR. Just that reviews will be slower. If you want you can split up into another PR also to make it more easy to handle, just ket me know order of review.. I guess it is toeplitz --> async-cufinufft --> async-gramoperator etc

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -640,24 +1087,28 @@ def _gram_op_sense_host(self, data, img_d):
...
-            data_batched *= smaps_batched
+            cp.multiply(data_single_d, smaps_batched, out=data_batched)
 
             self._gram_op_raw_device(data_batched, data_batched, padded_array)
-            for t, b in enumerate(idx_batch):
-                img_d[b, :] += data_batched[t] * smaps_batched[t].conj()
+            self._accumulate_coil_combine(img_d, i, data_batched, smaps_batched)
         img = img_d.get()
         img = img_d.reshape((B, 1, *XYZ))
         return img
```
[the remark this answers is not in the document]
paquiteau: hallucinated indentation. everything is fine here.

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -352,64 +388,267 @@ def op(self, data, ksp_d=None):
...
+            needs_norm = True
         else:
             op_func = self._op_calibless_host
-        ret = op_func(data, ksp_d)
+            needs_norm = False
 
-        ret /= self.norm_factor
+        ret = op_func(data, ksp_d)
+        if needs_norm:
+            ret *= 1.0 / self.norm_factor
```
chaithyagr, line 417: Why not just do this inside the op_func in each case?

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -352,64 +388,267 @@ def op(self, data, ksp_d=None):
...
+        ``anchor`` is the longest-lived object that owns the returned
+        memory: ``arr`` itself if it was already contiguous (no copy
+        needed), or the freshly made contiguous copy otherwise. Meant to
+        be passed straight to `_host_register`'s ``anchor=`` so repeat
+        calls with the same (contiguous) caller array hit the
+        registration cache instead of anchoring on a short-lived view.
+        """
+        contig = np.ascontiguousarray(arr)
+        anchor = arr if contig.ctypes.data == arr.ctypes.data else contig
+        return contig, anchor
```
chaithyagr: This is just a complex way to do ascontiguousarray right? The registration cache isnt really needed as numpy handles it i think and prevents copies.

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -352,64 +388,267 @@ def op(self, data, ksp_d=None):
...
+        Since ``n_trans`` is required to divide ``n_coils`` (see
+        `__init__`), a chunk never straddles a batch boundary: the batch
+        index for every one of the ``T`` coils in chunk ``i`` is the same
+        single value ``(i*T)//n_coils``, computed on the host as a plain
+        Python int (no device array/upload needed at all).
+        """
+        T = data_batched.shape[0]
+        vol = data_batched[0].size
+        b = (i * T) // self.n_coils
+        _coil_combine_kernel(data_batched, smaps_batched, b, T, vol, img_d, size=vol)
```
chaithyagr, line 523: I still feel its rather better to keep code simpler. what is the total gain with more complex code?

src/mrinufft/operators/interfaces/cufinufft.py
```diff
@@ -352,64 +388,267 @@ def op(self, data, ksp_d=None):
...
             if not self.smaps_cached:
                 smaps_batched.set(self.smaps[idx_coils].reshape((T, *XYZ)))
             else:
                 smaps_batched = self.smaps[idx_coils].reshape((T, *XYZ))
-            data_batched *= smaps_batched
+            # A chunk's T coils share a single batch (see `_coil_slice`):
+            # a single fused broadcast-multiply kernel straight from the
+            # caller's image view, instead of gathering (and copying) a
+            # redundant T-way repeat of it via fancy indexing.
+            cp.multiply(image_dataf[(i * T) // C], smaps_batched, out=data_batched)
```
chaithyagr, line 543: why not just use `*` than cp.multiply which makes it harder to read.

chaithyagr (commented):
Great job overall! I am so happy to have this up!  just some comments on code complexity

src/mrinufft/operators/interfaces/cufinufft.py
paquiteau: we fused the operation in the kernel so we avoid unnecessary copies, it's a bit faster as well :)

src/mrinufft/operators/interfaces/cufinufft.py
paquiteau: it saves memory, and we get 10-20% of improvement compared to the naive `a+=a * b.conj`. however the comment if overly verbose, let me fix this.

src/mrinufft/operators/interfaces/cufinufft.py
paquiteau: it ensure memory is properly managed (not creating new array or whatever)

PR #397, opened 2026-07-26
Read noise from twix data if it exists

chaithyagr:
The noise data is fairly small and worth to be added into header. Note that data is acquired at 5us, so you can alwayss calculate the noise_cov and scale the noise based on your needs in simulations (retrospective recon / SNAKE etc)


paquiteau:
Some merge conflict to fix, otherwise its a nice addition.

chaithyagr:
FIxed, will merge when green

paquiteau:
We have everything we need in #400 now, I think we can close this and use turbotwix, which is faster and better :)

